PR #6, opened 2014-03-22
Pdio 236 - fix bug 500 create user option parameters

roninio:


pferrel:
will not fix, add a feature request instead.

PR #36, opened 2015-01-12
Improve Java installation and locating in install.sh

pawel-n:
I have encountered some issues with the install.sh script on Gentoo linux. Installing Java thrown errors (apt-get...) and JAVA_HOME detection failed (because in Gentoo javac is a symlink to "/usr/libexec/eselect-java/run-java-tool.bash").

The script now:
* lets you decide if you want to install Java or not, and asks you for your distro if you do (currently only Debian/Ubuntu is supported),
* checks if JAVA_HOME is correct and in case of failure lets the user input the path manually,
* uses "read -e -p ..." which enables auto-completion of paths.

I have tested it only on Gentoo.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -18,6 +18,24 @@ USER_PROFILE=$HOME/.profile
 PIO_FILE=PredictionIO-$PIO_VERSION.tar.gz
 TEMP_DIR=/tmp
 
+# Ask a yes/no question, with a default of "yes".
+confirm () {
+  echo -ne $@ " [Y/n] "
+  read -r response
+
+  case $response in
+    [yY][eE][sS]|[yY]|"")
+      true
+      ;;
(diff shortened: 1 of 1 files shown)


dszeto:
Thanks a lot for the PR @pawel-n ! Before I could merge your PR, would you mind signing our CLA at http://prediction.io/cla first?

pawel-n:
I have signed it.

dszeto:
Thanks!

ramaboo:
Thanks @pawel-n Code looks great! I made a few wording changes and rewrote our other loop to use your confirm method which is really handy to have. I also moved a few things around to make sure our Docker install path remains interactive free.

I love how are script is getting better every day. If you have any more suggestions please let us know.  We will be rolling out these changes on install.prediction.io shorty!

PR #52, opened 2015-03-01
Import data into database before starting rails application

shekhargulati:
Currently, if someone is following this tutorial they will not see any comic as database is empty. They will see a loading spinner and error in their developer tool console. This gives a feeling that something is broken.


hitgirl:
Thanks!

PR #53, opened 2015-03-02
Fixed bug when create new app that does not return the key of the app

rolandobermudez:
Closes: #PDIO-557
Added initial code for train command
Updated doc in README.md


k4hoo:
Thanks!

PR #55, opened 2015-03-04
Guide how to blacklist recommended items

aparamonov:


docs/manual/source/templates/recommendation/blacklist-items.html.md
```diff
@@ -0,0 +1,140 @@
...
+
+```scala
+  def predict(model: ALSModel, query: Query): PredictedResult = {
+    // Convert String ID to Int index for Mllib
+    model.userStringIntMap.get(query.user).map { userInt =>
+      // create inverse view of itemStringIntMap
+      val itemIntStringMap = model.itemStringIntMap.inverse
+      // recommendProducts() returns Array[MLlibRating], which uses item Int
+      // index. Convert it to String ID for returning PredictedResult
+      val blackListedIds = query.blackList.map(model.itemStringIntMap.get(_).get) // NOTE: line added
```
k4hoo: Hi @aparamonov , need to handle the case that the blackListed item IDs not found in model (such as new item ID). eg.
val blackListedIds = query.blackList.map(model.itemStringIntMap.get(_)).flatten

aparamonov: Hi @k4hoo , good catch! I've just made a fix for that.

k4hoo:
Great doc and example! Would you merge to our "livedoc" branch instead? so it can go live! thanks!

aparamonov:
Kenneth, I've just created PR #61 on top of livedoc branch. Please merge this - I don't have sufficient privileges. Thanks!

PR #56, opened 2015-03-04
Bump Elasticsearch version to 1.4.4

pawel-n:
Hi,

I came across some dependency conflicts when working on a [template] (https://github.com/pawel-n/template-scala-parallel-sentiment-analysis) using [deeplearning4j-nlp] (http://mvnrepository.com/artifact/org.deeplearning4j/deeplearning4j-nlp/0.0.3.3). dl4j requires lucene 4.10.1, but Elasticsearch pulls in lucene 4.9. This results in a runtime crash.

I'd like to upgrade Elasticsearch library to version 1.4.4 to fix that problem. I have tested the patch using the engine I linked above.


dszeto:
Thanks for the patch. We are evaluating the impact to Elasticsearch 1.3 users and will get back to this as soon as possible.

dszeto:
Compatible with Elasticsearch 1.3 as a meta data store. Merging. Thanks!

PR #59, opened 2015-03-05
Fix link to the template gallery

detrevid:


hitgirl:
@detrevid Thank you for catching and fixing it!

PR #64, opened 2015-03-06
Added RateEvent to SimilarProductTemplate Under Examples

mkausik:


examples/scala-parallel-similarproduct/add-rateevent/src/main/scala/DataSource.scala
```diff
@@ -0,0 +1,114 @@
...
+      // targetEntityType is optional field of an event.
+      targetEntityType = Some(Some("item")))(sc)
+      // eventsDb.find() returns RDD[Event]
+      .map { event =>
+        val rateEvent = try {
+          event.event match {
+            case "rate" => RateEvent(
+              user = event.entityId,
+              item = event.targetEntityId.get,
+              t = event.eventTime.getMillis)
```
k4hoo: Hi, for "rate" event. I assume there will the "rating" property for the event. and it should be read here.

examples/scala-parallel-similarproduct/add-rateevent/src/main/scala/DataSource.scala
```diff
@@ -0,0 +1,114 @@
...
+      rateEvents = rateEventsRDD
+    )
+  }
+}
+
+case class User()
+
+case class Item(categories: Option[List[String]])
+
+case class RateEvent(user: String, item: String, t: Long)
```
k4hoo: Hi, the RateEvent class should have "rating" field to store the rating value

mkausik: Added "rating" field to RateEvent class.

examples/scala-parallel-similarproduct/add-rateevent/src/main/scala/DataSource.scala
mkausik: I have fixed the issue.

examples/scala-parallel-similarproduct/add-rateevent/src/main/scala/ALSAlgorithm.scala
```diff
@@ -0,0 +1,236 @@
...
+
+        if (uindex == -1)
+          logger.info(s"Couldn't convert nonexistent user ID ${r.user}"
+            + " to Int index.")
+
+        if (iindex == -1)
+          logger.info(s"Couldn't convert nonexistent item ID ${r.item}"
+            + " to Int index.")
+
+        ((uindex, iindex), 1)
```
k4hoo: Hi @mkausik,  for algorithm, should use rating as well (instead of 1), please see https://github.com/PredictionIO/PredictionIO/blob/2d30354febb7fc6b46df9c9869fd33a5cb2917cf/examples/scala-parallel-ecommercerecommendation/train-with-rate-event/src/main/scala/ALSAlgorithm.scala#L94 as example.

k4hoo:
Thanks @mkausik !

PR #71, opened 2015-03-09
Fix core to be consistent with scalastyle

jtescher:
Mostly public method types, spaces after comments, and multi-line if statements


dszeto:
Thanks!

PR #73, opened 2015-03-11
Example of recommendation template with filtering by category

ArturSoler:
Recommend items, limiting results to those belonging to a category specified by the query.

A detailed explanation of changes can be found in the file README.md

### DIFF
--- a/examples/scala-parallel-recommendation/filter-by-category/README.md
+++ b/examples/scala-parallel-recommendation/filter-by-category/README.md
@@ -1,0 +1,239 @@
+# Recommendation Template With Filtering by Category
+
+This engine template is based on the Recommendation Template version v0.1.2. It's modified so that each item has a set
+of categories, queries include a `category` field and results only include items in the given category.
+
+## Documentation
+
+Please refer to http://docs.prediction.io/templates/recommendation/quickstart/
+
+## Development Notes
+
+### Sample data
+
+This example uses two different files from the MovieLens 100k data. [Download](http://files.grouplens.org/datasets/movielens/ml-100k.zip)
+
+The first file is the ratings data file, which is a tab separated file. The first 3 fields are used: user id, item id
+and rating.
+
+The second file is the items data file, in where different fields are separated by the `|` character. The fields of
+interest are the first one (item id) and 19 category fields, starting at the 6th one and going till the end. Each
+of these fields has the value `1` if the item is the given category, and `0` otherwise.
+
+### Importing sample data
+
+The `data/import_eventserver.py` script allows importing the sample data. Valid arguments are:
+
+* `--url`: Event server URL
+* `--access_key` App access key
+* `--ratings_file`: Ratings file path
+* `--items_file`: Items file path
+
+### Changes to Engine.scala
+
+Added `category` field to the `Query` class:
+
+```scala
+case class Query(
+  user: String,
+  num: Int,
+  category: String
+) extends Serializable
+```
+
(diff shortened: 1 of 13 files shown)


k4hoo:
Hi @ArturSoler,
creating a separate model for each category is not ideal solution because
- duplicates many model if many categories
- should also support query by multiple categories, for example,
{"user": "1", "num": 4, "categories": ["action", "adventure"]})
Some reference you can take a look:
http://docs.prediction.io/templates/recommendation/blacklist-items/
http://docs.prediction.io/templates/similarproduct/quickstart/
https://github.com/PredictionIO/template-scala-parallel-ecommercerecommendation

Thanks!

ArturSoler:
Hi @k4hoo,

I understand the issue and have updated the example to use a single model. I have also updated the `Query` to support multiple categories, as you suggested.

Thanks for your pointers, they have made this much easier. Let me know if there's any other issue.

Artur

k4hoo:
Hi Artur,

thanks for making the changes.
But creating a Map of category to a List of items seem to duplicate and waste storage space. Possible to store the item category for each item instead?
instead of,

```scala
val categoryItemsMap: Map[String, Set[Int]])
```
use something like:

```scala
case class Item(categories: Option[List[String]])

val items: Map[Int, Item]
```
Then in predict(), find out list of item that belong to the categories in query.

How do you think?

Thanks
Kenneth

ArturSoler:
Hi Kenneth,

I've computed the storage and time cost for both alternatives and I've come up with the following:

**Current approach**

* Storage: There is a `Map` for each different category, plus an `Int` for each match between a category and an item. Total: `O(categories + items * average(categories per item) * size(Int))`

* Time: There's a set lookup (effectively constant time) for each considered item and for each queried category. Total: `O(considered items * queried categories)`

**Alternative approach**

* Storage: There is an `Item` and a `List` object for each item, plus a `String` for each match between a category and an item. Total: `O(items + items * average(categories per item) * size(String))`  

* Time: For each considered item, there's a check between each of the item categories and each of the query categories. Total: `O(considered items * average(categories per item) * queried categories)`

The storage cost in the alternative approach would be a bit higher, since a `String` takes more space than an `Int`. This could be solved by mapping category `String`s to `Int`s as it's done for user and item ids. The number of items is probably going to be in general higher than the number of categories, but this part of the expression is not going to dominate except in very sparse sets of data.

Regarding time, the alternative approach would also be a bit more costly, but this could also be solved by substituting the `List` of categories by a `Set` of categories.

With these two optimizations, both storage and time cost would be equivalent between the two alternatives. 

I've thought about another possible storage optimization, which would not help in the asymptotic case but would probably help in most real-life cases. That would be, in the alternative approach, to try to have the JVM reuse the `List` or `Set` of item categories where two items have the same set of categories. E.g., if there are two films that are both "action" and "adventure", in the naive implementation there would be 2 `List("action", "adventure")` objects (with the possible `Set` and `Int` variants), while we could detect those cases and reuse a single object. Testing on the 100k MovieLens data, from 1682 total items there are only 216 different sets of categories. In other domains with more categories, though, the benefits of this approach would be less.

Please let me know which variation do you prefer to have implemented.

Kind regards,
Artur

k4hoo:
Hi Artur,

Great analysis and agree with you!
Let's keep your current implementation.

Thanks!
Kenneth

k4hoo:
Thanks @ArturSoler! Excellent work!

PR #74, opened 2015-03-12
confirm delete existing tmp file before download

ngheungyu:
I encountered an error before where interrupting the tar download would cause subsequent install attempts to fail. Deleting the tmp file would make sure it gets download again.


dszeto:
Looking good! Would you mind applying the change also to all other downloaded files (Spark, Elasticsearch, and HBase), and submit the PR against the master branch instead? That way the change will be immediately deployed live to https://install.prediction.io/install.sh. Thanks!

ngheungyu:
how about i silently delete those files?


Best Regards,
Nicholas

PR #75, opened 2015-03-12
Delete tmp files in case corrupted

ngheungyu:
I encountered an error before where interrupting the tar download would cause subsequent install attempts to fail. Deleting the tmp file would make sure it gets download again.


dszeto:
It would be nice to retain the cache functionality, because it is useful during development not needing to re-download large files every time.

ngheungyu:
ok, i will add 4 questions for that then


Best Regards,
Nicholas

ngheungyu:
added, please review.

dszeto:
Looking good and tested on our side as well. Thanks!

PR #79, opened 2015-03-15
Example of filtering recommendations on custom item property.

mkorolyov:
Often people want to filter out items created before X date in their recommendation.

I have customized the engine to filter the result set on creationYear parameter from query.

All additional changes are marked with `HOWTO` comments.


k4hoo:
Hi Maxim,

Thanks!
- Which template and which version is your modification based on? Would you update README about this?

- The Serving has code for combining multiple algorithm's PredictedResult. For simplicity and the purpose of this example, i think better to make the Serving simple as original (such as https://github.com/PredictionIO/template-scala-parallel-recommendation/blob/v0.1.2/src/main/scala/Serving.scala). so that it won't confuse user that the Serving change is needed for this.

- There is a Scala import program inside src/. Do you think it's cleaner if make it a separate project in another directory (say, a data/ directory ) so that not mixing with the engine source code ? And would you update REAMDE to teach user how to run the import program too.

Thanks
Kenneth

mkorolyov:
Hi Kenneth,

Thanks for your comments, i have simplified the Serving. 
I have moved data import script under `./data` in the example.
I have also updated `README.md` with instructions on how to use the import command line tool and pointed template version that i was based on. 
Let me know if there's any other things i should improve.

Thanks
Maxim.

k4hoo:
Hi Maxim,

Thanks for the updates. Nice work!

- Could you describe a bit about what and why Serving is modified in README as well?
(e.g re-sort item based on the year?).

- In Serving, is there any reason that you do groupBy item id and sum the scores (i thought the result.itemScores should contain top unique items already?)

```scala
result.itemScores
        .groupBy(_.item)
        .mapValues { itemScores ⇒
          itemScores.map(_.score).reduce(_ + _) ...
```

Thanks
Kenneth

mkorolyov:
Kenneth,

excess grouping code was accidentally left after Serving simplifying. Sorry for that, my bad. It is fixed now. I have added actual recommendation result filtering as a `bonus` like Hagay Gazuli mentioned in the [google group](https://groups.google.com/forum/#!searchin/predictionio-user/created$20%7Csort:date/predictionio-user/LEHxuc0Bu_0/W9RkAApvivsJ)

Thanks
Maxim

k4hoo:
haha I see!  Like your bonus feature! @mkorolyov   
Thanks for updating README to clarify that.

PR #83, opened 2015-03-22
Install from source build and config instruction tweaks

pferrel:
BTW why are their both hbase conf and root variable in pio-env.sh? Seems like conf can be derived from the root. Conf should be kept with the HBase installation directory don't you think? That way the binary and config can be changed independent of PIO. HBase can then be treated as independent of PIO by experts and still works when bundled.

Following the HBase install quickstart leads one to believe the conf is with HBase, not in pio/conf. So there is confusion where you tell the user to edit the hbase/conf but the edits are ignored by pio-start-all since it finds the hbase conf at pio/conf.

The instructions will create a working installation but this seems a bit inconsistent.

### DIFF
--- a/docs/manual/source/install/install-sourcecode.html.md.erb
+++ b/docs/manual/source/install/install-sourcecode.html.md.erb
@@ -68,10 +68,13 @@ Elasticsearch.
 
 If you are not using the default setting at localhost. You may change the following in ```conf/pio-env.sh``` to fit your setup.
 
+Make sure to set PIO_STORAGE_SOURCES_ELASTICSEARCH_HOME
+
 ```
 PIO_STORAGE_SOURCES_ELASTICSEARCH_TYPE=elasticsearch
 PIO_STORAGE_SOURCES_ELASTICSEARCH_HOSTS=localhost
 PIO_STORAGE_SOURCES_ELASTICSEARCH_PORTS=9300
+PIO_STORAGE_SOURCES_ELASTICSEARCH_HOME=/path/to/elasticsearch
 ```
 
 #### <a name="hbase"></a>HBase Setup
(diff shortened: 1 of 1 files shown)


dszeto:
Largely historical. `$HBASE_CONF_DIR` set to `$PIO_HOME/conf` has been around since the beginning of 0.8, and is included in the JVM classpath for the storage layer's HBase client. This is why it seems incomplete, as only the quorum configuration is required if you have a setup connecting to a remote HBase cluster.

`$PIO_STORAGE_SOURCES_HBASE_HOME` was recently added for the `bin/pio-start-all` script. Agree that `HBASE_CONF_DIR` can be derived from that variable if we always require `$PIO_STORAGE_SOURCES_HBASE_HOME` to be set. `bin/pio-start-all`, however, is designed for quick-and-dirty standalone setups, so `$PIO_STORAGE_SOURCES_HBASE_HOME` does not necessary need to exist for distributed setups. Despite the fact that they serve different purposes, I agree that the current documentation is not clear and makes them kind of confusing.

So in summary, the current situation is due to some hacking to handle all of the following cases:
- PIO with remote HBase
- PIO with bundled local HBase
- PIO with non-bundled local HBase

I do not have a better solution at the code level at the moment, besides improving the documentation. Any suggestion is welcome.

Thanks for the PR! Merging the docs now.

pferrel:
Yes they do work for more standard or networked HBase. However I'd suggest that PIO have only environment variable pointing to hbase (though the name is somewhat verbose and confusing Maybe just PIO_HBASE_HOME=)

    PIO_STORAGE_SOURCES_HBASE_HOME=/path/to/hbase

inside of the hbase dir you'll find "conf" for all three cases. This may mean changing your job/driver code but will rationalize the hbase conf to be the same for all cases; "edit in your hbase/conf/hbase-site.xml" This will handle all your situations

* PIO with remote HBase
* PIO with bundled local HBase
* PIO with non-bundled local HBase

Further it will allow users to follow the HBase instructions and tutorials verbatim.

pferrel:
Actually the basis of the suggestion is to remove hbase-site.xml from pio/conf and leave it with hbase/conf

dszeto:
Agree that `conf/hbase-site.xml` has been causing some confusion with our team as well. Also, apparently HBase is using `$HBASE_CONF_DIR` as well so changing it to other values, and then calling `start-hbase.sh` inside `pio-start-all` will screw things up.

For the remote HBase case, I think it is categorized as advanced usage and some documentation should clear it up.

PR #84, opened 2015-03-22
[pio-shell] interactive PIO shell

whhone:
Add a script to start the PIO shell. There are two modes to start the shell.

1) `pio-shell`:
Start the pio shell by sbt. It includes the pio library.

2) `pio-shell --with-spark`
Start the pio shell by spark-shell. It includes both pio library
and the Apache Spark. It requries the Apache Spark setup and also takes
longer to start.

For example, https://gist.github.com/whhone/016088d3c3dfd73aaf9f

### DIFF
--- a/bin/pio-shell
+++ b/bin/pio-shell
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Copyright 2015 TappingStone, Inc.
+#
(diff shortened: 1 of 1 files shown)


dszeto:
Thanks! That's very useful.

PR #86, opened 2015-03-25
Example of E-commerce recommendation with weighted items

ArturSoler:
Example based on E-Commerce Recommendation Template with a modification so that items can be given weights that modify the likeliness with which they are recommended.


k4hoo:
Great work! Thanks @ArturSoler !

PR #91, opened 2015-04-01
Code cleanup in core

detrevid:


dszeto:
Appreciate your effort. One quick minor note, would you please dial up "Class count to use import with '_'" to some large number (assuming you're using IntelliJ)? We intentionally try to make import statements as specific as possible.

dszeto:
My apology. Figured out nothing has been collapsed. Will merge this shortly.

dszeto:
Looking good and merged. Thanks!

PR #92, opened 2015-04-01
Crossval

jglidden:


e2/src/test/scala/io/prediction/e2/evaluation/CrossValidationTest.scala
```diff
@@ -0,0 +1,124 @@
+package io.prediction.e2.evaluation
+
+import io.prediction.e2.evaluation.CrossValidationTest.EmptyEvaluationParams
```
yipjustin: Probably don't need this as you have `CrossValidationTest.EmptyEvaluationParams()` in the code

e2/src/main/scala/io/prediction/e2/evaluation/CrossValidation.scala
```diff
@@ -0,0 +1,59 @@
...
+  def splitData[D: ClassTag, TD, EI, Q, A](
+
+     evalK: Int,
+     dataset: RDD[D],
+     evaluatorInfo: EI,
+     trainingDataCreator: RDD[D] => TD,
+     queryCreator: D => Q,
+     actualCreator: D => A): Seq[(TD, EI, RDD[(Q, A)])] = {
+
+    val indexedPoints = dataset.zipWithIndex
```
yipjustin: Just FYI. No need to fix.

`RDD.zipWtihIndex()` returns a continuous segment of index, from 0 until size. There is also a function `RDD.zipWithUniqueId()` returns a indexed sequence but there can be gaps in between. The latter is more efficient as it doesn't require a shuffle step, the former has a higher chance of attaining uniform distribution across all testing data size, i.e. `"Testing data size" should  "be within 1 of total / evalK"`.

jglidden: Thanks that's helpful to know. I'll keep that in mind if this ends up being a performance bottleneck.

e2/src/test/scala/io/prediction/e2/evaluation/CrossValidationTest.scala
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+
+  "Testing data size" should  "be within 1 of total / evalK" in {
+    val labeledPointsRDD = sc.parallelize(labeledPoints)
+    val splits = evalKs.map(k => k -> splitData(k, labeledPointsRDD))
+    val diffs = splits.map { case (k, folds) =>
+      folds.map(fold => fold._3.count() - dataCount / k)
+    }
+    diffs.flatMap(folds => if (folds.exists(_ > 1)) Some(true) else None) should be('empty)
```
yipjustin: Two things.

1. need to handle negative value, i.e. `< -1`?
2. you are trying to find if all values in the nested sequence `diffs` passes certain condition. To make it more scalatest-ish, You can:  `forAll(diffs) { diff => forAll(diff) { e => math.abs(e) should be < 1 } }`

jglidden: Great, I wasn't sure of the idiomatic way to test that all elements of a sequence meet a certain criteria.

e2/src/test/scala/io/prediction/e2/evaluation/CrossValidationTest.scala
```diff
@@ -0,0 +1,124 @@
...
+    diffs.map(folds => folds.sum) should be(evalKs.map(k => dataCount % k))
+  }
+
+
+
+  "Training + testing" should "equal original dataset" in {
+    val labeledPointsRDD = sc.parallelize(labeledPoints)
+    val splits = evalKs.map(k => splitData(k, labeledPointsRDD))
```
yipjustin: Similarly, we want to test overall evalK, and each element in evalK are used independently. The code can be
```
forAll(evalKs) { k => 
  val split = splitData(k, labeledPointsRDD)
  forAll (split) { fold => 
    val (training, testing) = toTestTrain(fold)
    (training ++ testing).toSet should be labeledPoints.toSet
  }
}
```

yipjustin:
Hi jglidden,

LGTM.

There are some minor things in the test code to make it more scalatest-ish. Will merge it once they are fixed.

Thanks.

Justin

yipjustin:
thx!

PR #93, opened 2015-04-08
Example engine template to recommend users

agarella:
An engine template example based on the similar product template to recommend users.
The recommendation is based on *User views User* events.


k4hoo:
hi @agarella, thanks! great work.
would you add the following to the README:

- an overview to describe what this example does?
- which version of original template it's based on ?
- explain the modification and difference from the original template.

an example of the README would be:
https://github.com/PredictionIO/PredictionIO/tree/develop/examples/scala-parallel-similarproduct/add-and-return-item-properties

Thanks
Kenneth

agarella:
Hi Kenneth,

I have updated the README.

k4hoo:
thanks!

PR #96, opened 2015-04-17
Minimum basket size parameter description

pawel-madurski:
This is part of documentation describing minBasketSize algorithm property


k4hoo:
Thanks @pawel-madurski for your idea!

PR #99, opened 2015-04-21
MailChimp Webhook Connector 

gracexu94:
Added connector code and tests for MailChimp webhooks, for all the events listed in the MailChimp API (https://apidocs.mailchimp.com/webhooks/).

### DIFF
--- a/data/src/main/scala/io/prediction/data/webhooks/mailchimp/MailChimpConnector.scala
+++ b/data/src/main/scala/io/prediction/data/webhooks/mailchimp/MailChimpConnector.scala
@@ -255,12 +255,11 @@ private[prediction] object MailChimpConnector extends FormConnector {
 
     val json =
       ("event" -> "cleaned") ~
-      ("entityType" -> "user") ~
-      ("entityId" -> data("data[campaign_id]")) ~
-      ("targetEntityType" -> "list") ~
-      ("targetEntityId" -> data("data[list_id]")) ~
+      ("entityType" -> "list") ~
+      ("entityId" -> data("data[list_id]")) ~
       ("eventTime" -> eventTime) ~
       ("properties" -> (
+        ("campaignId" -> data("data[campaign_id]")) ~
         ("reason" -> data("data[reason]")) ~
         ("email" -> data("data[email]"))
       ))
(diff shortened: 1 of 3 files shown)


data/src/main/scala/io/prediction/data/webhooks/mailchimp/MailChimpConnector.scala
```diff
@@ -97,4 +107,200 @@ private[prediction] object MailChimpConnector extends FormConnector {
...
+      ("entityType" -> "user") ~
+      ("entityId" -> data("data[campaign_id]")) ~
+      ("targetEntityType" -> "list") ~
+      ("targetEntityId" -> data("data[list_id]")) ~
+      ("eventTime" -> eventTime) ~
+      ("properties" -> (
+        ("reason" -> data("data[reason]")) ~
+        ("email" -> data("data[email]"))
+      ))
+
```
k4hoo: modeling cleaned event with "user" as entityType and compaign_id as entityId seems not quite right. any better way?

data/src/main/scala/io/prediction/data/webhooks/mailchimp/MailChimpConnector.scala
```diff
@@ -97,4 +107,200 @@ private[prediction] object MailChimpConnector extends FormConnector {
...
+    "data[list_id]": "a6b5da1054"
+    */
+
+    // convert to ISO8601 format
+    val eventTime = Utils.dateTimeToString(parseMailChimpDateTime(data("fired_at")))
+
+    val json =
+      ("event" -> "campaign") ~
+      ("entityType" -> "user") ~
+      ("entityId" -> data("data[id]")) ~
```
k4hoo: the data[id] is the compaign id, so the "entityType" make more sense to be "campaign" as well?

gracexu94: Thanks Kenneth! I changed the "entityType" for the campaign sending status
to "campaign" like you suggested.
For the cleaned email, since it seems automatic on the part of Mailchimp
(at least according to their docs
<http://kb.mailchimp.com/delivery/deliverability-research/soft-vs-hard-bounces>),
would it make more sense to model it with "entityType" = "campaign" and
keep the "entityId" as the "campaign_id"?

Or perhaps make the "entityType" = "list", the "entityId" the "list_id" and
put "campaign" and "campaign_id" as the "targetEntityType" and
"targetEntityId"?

Thanks again!

k4hoo: Thanks Grace. 
for "cleansed" event, How about also use "list" as entityType without targetEntityType?
ie. EntityType="list" has "cleaned" event with the email and campaign_id inside the properties.

k4hoo:
Thanks @gracexu94 ! great work!

PR #100, opened 2015-04-23
Added the ability to set the elasticsearch cluster name

darkseed:
The current elasticsearch driver does not allow, clusters other that the default 'elasticsearch', this works fine on a local machine, but presents problems in a clustered environment, when you run multiple clusters.

Added the ability to specify the elasticsearch cluster name, with a fallback to the default 'elasticsearch' if not set.


dszeto:
Although this is already covered by ES_CONF_DIR, this is a common case and is very handy. Thanks @darkseed !

PR #102, opened 2015-04-24
Changed events.json into my_events.json

Jeroenvos:
In the python sdk part, the script writes to my_events.json. In the example line below events.json is used.
When learning PredictionIO by copy-pasting examples without really getting what one is doing this might be confusing (That is the way I learned it in the first place :-)).
So would like to I suggest to be consistent in the SDK and line examples in this case.


k4hoo:
Good suggestion! Thanks @Jeroenvos !

PR #103, opened 2015-04-29
Updated MailChimp Webhook

gracexu94:
Made the [INTERESTS] field in the Mailchimp Webhook connector optional to match actual webhook data.


k4hoo:
Thanks Grace!

PR #107, opened 2015-05-03
Enabling unit test in travis CI

ohenak:
* enable travis for `develop` branch
* use JVM from travis instead of Ruby VM
* prepare spark, postgre and hbase for pio
* build doc
* run unit test using built-in sbt (current is 0.13.7)


dszeto:
Looking good. Thanks!

PR #110, opened 2015-05-05
Call jps from JAVA_HOME if JAVA_HOME is set

ch33hau:
I'm using Fedora with Oracle Java that installed manually. 
During **pio-start-all**, I will see below error message:
> PredictionIO/bin/pio-start-all: line 28: jps: command not found

For my case, the **jps** is available in *JAVA_HOME/bin/jps* but not in *PATH*, unless configure **alias** or **update-alternatives** for **jps** manually. 

Suggest to call **jps** from *JAVA_HOME/bin* if *JAVA_HOME* is set, else remain to current behavior.


dszeto:
Looks good to me. Thanks!

PR #111, opened 2015-05-05
Add support for serializing lists created via Arrays#asList(Object...)

detrevid:


dszeto:
Thanks! Any specific use case by chance?

detrevid:
Yeah. I am using other library with PIO and wanted to store a class in my model that can internally hold a list created via Arrays#asList. However, there was a new commit to the library and I do not need it any more. 
Anyway, I can add more kryo serializers now or in the future if needed (so less people will get "Failed to invert" error while deploying an engine).

pferrel:
Is this still an issue? Can we resolve?

pferrel:
@dszeto is the KryoInstantiator a non-core Kryo add-on? Don't know Kryo well enough but since it's at the core of a lot of very hard to find errors I try to avoid touching it and using Scala/Java serialization even though it may be slightly slower. Do we have a POV on mucking with Kryo? Should this be merged?

pferrel:
No clear reason for this and remaining conflict seem like a wontfix

PR #113, opened 2015-05-06
Enabling code coverage using scoverage

ohenak:
* `sbt/sbt clean coverage test` to instrument then collect coverage data from all unit tests
* current coverage is 1.62% in total, need more unit tests
* switch travis to use container-based infra for better RAM allocation (potentially solve the problem with time out jobs)


dszeto:
Thanks!

PR #115, opened 2015-05-11
Update intellij.html.md.erb

pferrel:
if there are merge issues, only the part after `### Executing a query` is new.


dszeto:
Thanks!

PR #119, opened 2015-05-24
Fix typo

mscoutermarsh:
:star2:


k4hoo:
Thanks @mscoutermarsh ! good catch!

PR #129, opened 2015-07-21
Fix a bug due to an invalid column name when using JDBC as data source

minhtule:
When testing the Event Server with `data/test.sh` using PostgreSQL as datasource, the last test failed with the following message error.

```
400 Bad Request
Server: spray-can/1.3.2
Date: Tue, 21 Jul 2015 12:05:33 GMT
Content-Type: application/json; charset=UTF-8
Content-Length: 108

{"message":"org.postgresql.util.PSQLException: ERROR: column \"starttime\" does not exist\n  Position: 303"}
expect 200
```


dszeto:
Good catch. Thanks @minhtule !

PR #132, opened 2015-07-22
Batch requests support

minhtule:
An attempt to implement support for batch requests that was discussed on the [development google group] (https://groups.google.com/forum/#!topic/predictionio-dev/Fd_Dt2Lhvy4).
The specification can be found [here](https://docs.google.com/document/d/1byXQ7eM1pJ1g8O4hVBvEi1AuMSCx7osSrJk3FKwAaz4/edit). 
@dszeto You mentioned that I should create tasks on JIRA but I'm not sure what to create so please advise if I still need to.

### DIFF
--- a/data/src/main/scala/io/prediction/data/api/EventServer.scala
+++ b/data/src/main/scala/io/prediction/data/api/EventServer.scala
@@ -30,6 +30,7 @@ import io.prediction.data.storage.Channels
 import io.prediction.data.storage.DateTimeJson4sSupport
 import io.prediction.data.storage.Event
 import io.prediction.data.storage.EventJson4sSupport
+import io.prediction.data.storage.BatchEventsJson4sSupport
 import io.prediction.data.storage.LEvents
 import io.prediction.data.storage.Storage
 import org.json4s.DefaultFormats
@@ -46,6 +47,7 @@ import spray.routing.authentication.Authentication
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.util.{Try, Success, Failure}
 
 class EventServiceActor(
     val eventClient: LEvents,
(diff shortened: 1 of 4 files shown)


dszeto:
I have created a ticket for you to keep track of this: https://predictionio.atlassian.net/browse/PDIO-642

Looks good to me. The next iteration could use some refactoring with the single event creation code. Thanks!

minhtule:
That was fast! Thanks for creating the ticket. I didn't know it was that simple >.<

I tried to reuse the single event creation code but the return types are different so I was stuck. For single event, the completion handler returns `(status, Map("message": "..."))` while each event in the batch request returns `Map("status": 400, "message": "...")`. Do you have any tips on how to refactor this?

Btw, should I update the documentation page to reflect this support for batch requests?

pferrel:
Looks like this creates a new REST endpoint? is there a reason to not include this in any event? So any event can have either a single object or array of objects?

minhtule:
Yes, this creates a new REST endpoint. We could handle both batch and single event requests at the same end point but in that case, both the client and the server handler need to do extra checking to know  what type of request/response it is dealing with. Because the format for batch requests and responses are quite different from those of single event (e.g. failure of creating some single event in a batch request will not response with an error code unlike a single request), I think it is better to separate them.

pferrel:
But why would you not respond with an error if one of the batch failed? The use case I have in mind is where many events are created with a single user action. This shouldn't require many events to be sent separately so if we use the batch endpoint we would never know when a failure happened.

Ideally the client should get an error so they can retry, right? It might not be usable without the error return for my usecase.

Also the format should be a json event or a json array of events, right? This seems easy to check for in a single endpoint, where array = batch, single = one event.

We could have them use the batch endpoint always, creating an array of one or more events with each request but we would need the error as much for batch as for single requests.

minhtule:
For batch request, the response's status code is meant for system failure (e.g. server down or database connection error). Even if we return an global error code when a single event failed to create, we still have to go through all events to check which exact event fails in order to retry. So in batch request, each event will have its corresponding status code in the response's body (please see the sample response in the doc).

> Also the format should be a json event or a json array of events, right? This seems easy to check for in a single endpoint, where array = batch, single = one event.

Yes, this is true. The server can check the request format to see if it is a batch or single request. But then the client also has to check the response format to determine if it is handling a batch or single request. I think this creates unnecessary work. Of course, one can always send a batch request with only 1 event though.

pferrel:
ok this looks good to me. The errors reported from the endpoint are better than I'd expect. Several of our users will probably use this endpoint exclusively so I can't wait to try it, Thanks Minh!

minhtule:
Yes! (to all 3 questions)

PR #145, opened 2015-09-17
Fix install.sh for Docker mode (Elastic Search + Hbase not installed)

clemp6r:
The `source_setup` variable is not defined in Docker mode, so no database is installed.


pferrel:
@clemp6r no effect on other install modes?

clemp6r:
Nope, check the diff, it's just one line that was missing in the "if (docker)" section.

PR #147, opened 2015-09-20
Change repo order, and follow redirect with curl

tuxdna:
Fix fo following issue:

https://github.com/PredictionIO/PredictionIO/issues/146

### DIFF
--- a/sbt/sbt-launch-lib.bash
+++ b/sbt/sbt-launch-lib.bash
@@ -37,8 +37,8 @@ jar_file () {
 
 acquire_sbt_jar () {
   SBT_VERSION=`awk -F "=" '/sbt\\.version/ {print $2}' $sbt_bin_dir/../project/build.properties`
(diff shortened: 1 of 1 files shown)


pferrel:
A lot of people getting this error, @tuxdna ok to pull?

BTW good to see you here.

pferrel:
@dszeto any opinion on this? other than adding verbose to curl and wget he changed the order of repos to search and told curl and wget to follow redirects. 

haven't used this version though

tuxdna:
The real issue here is that the repo "typesafe.artifactoryonline.com" doesn't contain a valid "sbt-launch.jar".

And, the second repo "repo.typesafe.com", does have a valid "sbt-launch.jar", but then it requires wget/curl to follow redirects.

This exactly what this patch fixes.

pferrel:
lot of people seeing this so merging and will test

PR #154, opened 2015-10-01
Update train-with-rate-event.html.md.erb

cristiancrc:
- typo in preparator.scala filename
- update engine file name according to latest scala ecommerce template
- focus attention on parameter to be removed when switching to explicit training


k4hoo:
Thanks @cristiancrc  for improving the doc!

PR #165, opened 2015-10-27
Fix JDBC driver issue w/ MySQL and PostgreSQL

EmergentOrder:


pferrel:
Have you checked this with the UR? The config there is jdbc for all stores (metadata etc) but ES configured as a generic base storage. Smalled installations may want this instead of HBase + ES

EmergentOrder:
Works with UR, with custom conf, which I'll add to the UR docs.

pferrel:
Great, thanks

PR #169, opened 2015-10-28
Bad link corrected to reflect correct URL

vaer-k:


tuxdna:
Links are alright now.

PR #173, opened 2015-10-28
Add direction for how to respond to prompt for name

vaer-k:
No information is provided in this document to guide users in their response to this prompt. Copying this modified text over from the way it appears on https://docs.prediction.io/machinelearning/dimensionalityreduction/


vaer-k:
Looks like my suggestion is mistaken. There is a need for some guidance here though.

PR #180, opened 2015-11-10
Update deprecated zeppelin link to incubator

y3nk0:
The existing github installation can possibly ouput errors with incompatible version 0.0.20 of frontend-maven-plugin. Since it is also deprecated, it is better to change the link to the new one: https://github.com/apache/incubator-zeppelin.git


pferrel:
I you update the PR to point to the `livedoc` branch it will get auto published, and thanks.

PR #181, opened 2015-11-11
Update deprecated zeppelin link to incubator link

y3nk0:
The existing installation can possibly output errors with incompatible version 0.0.20 of frontend-maven-plugin. Since it is also deprecated, it is better to change the link to the new one: https://github.com/apache/incubator-zeppelin.git


pferrel:
Thanks

PR #183, opened 2015-11-20
Change data type of event table to varchar(255)  when creating index.

### LINKED ISSUE #159: Could I change data type of JDBC event table?
I use mysql as a event table for some reasons.
LEventStore.findByEntity function with entityType and entityId is slow and sometimes I am getting java.util.concurrent.TimeoutException.
I want to create index on this columns to improve performance.

https://github.com/PredictionIO/PredictionIO/blob/65a64c902ad7671dfb2fb0e29d8af806a01c55d0/data/src/main/scala/io/prediction/data/storage/jdbc/JDBCLEvents.scala#L43-L56

So, I modified these columns after creating table in my scripts such as below.

```
alter table pio_event_1_1 modify column entityType varchar(10) not null;
alter table pio_event_1_1 modify column entityId varchar(255) not null;
create index pio_event_1_1_idx1 on pio_event_1_1 (entityId);
create index pio_event_1_1_idx2 on pio_event_1_1 (entityType);
```

Is there any better way for this situation?

schon:
Improvement performance when findByEntity is called.

Closes #159

### DIFF
--- a/data/src/main/scala/io/prediction/data/storage/jdbc/JDBCLEvents.scala
+++ b/data/src/main/scala/io/prediction/data/storage/jdbc/JDBCLEvents.scala
@@ -38,9 +38,38 @@ class JDBCLEvents(
   implicit private val formats = org.json4s.DefaultFormats
 
   def init(appId: Int, channelId: Option[Int] = None): Boolean = {
+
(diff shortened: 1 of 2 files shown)


EmergentOrder:
Tested, works, merged, thanks @schon

PR #184, opened 2015-12-01
Update dase.html.md.erb

ic:
Update a sample feature data to versions above 0.3.1, and add a warning stolen from the quick start guide.


k4hoo:
Thanks @ic

PR #186, opened 2015-12-03
C# Lib has been updated to support 0.9.4

dgilling:


pferrel:
We are on 0.9.5 now, any reason this wouldn't apply?

dgilling:
No should be fine on 0.9.5. I don’t see any API differences, although haven’t tested on 0.9.5
I updated lib awhile back but noticed the notes where out of date today.

Derric

PR #191, opened 2015-12-11
Add better error reporting to pio import

### LINKED ISSUE #137: Add better error reporting to pio import
catch parsing errors when importing in parallel from text/json or other file types. Often a single malformed line will be very difficult to find so log the malformed line before throwing an abort exception.  Consider exception counting for parsing, allow some small number before aborting, this will catch a blank line at the end and simple exceptions.

schon:
Closes #137


EmergentOrder:
Looks good, thanks @schon

PR #192, opened 2015-12-22
Delete application data in non-interactive mode

### LINKED ISSUE #149: Delete application data in non-interactive mode
Please provide a way to delete application data without user interaction, so this can be automated (i.e in bash scripts). E.g.:

    pio app data-delete -y MyApp

schon:
Add --yes, -y option when deleting app and channel.
It is useful to execute a batch script

Closes #149

### DIFF
--- a/tools/src/main/scala/io/prediction/tools/console/App.scala
+++ b/tools/src/main/scala/io/prediction/tools/console/App.scala
@@ -25,6 +25,7 @@ case class AppArgs(
   channel: String = "",
   dataDeleteChannel: Option[String] = None,
   all: Boolean = false,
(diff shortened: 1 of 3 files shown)


pferrel:
Looks good. Can we have a --force CLI instead of --yes, More self-explanatory

schon:
@pferrel Thanks for your comment. I totally agree with you to use "--force". What do you think about short option "-y". Do you think we don't need a short option or change "-f" in accordance with "--force"?

pferrel:
yes to -f.

schon:
@pferrel Modified it. Could you please check this PR. Thanks.

PR #194, opened 2015-12-27
Update deploy-cloudformation.html.md.erb

markbar:
Updating the doc to reflect user experience in following current documentation. http://stackoverflow.com/questions/32408380/deploying-predictionio-with-aws-cloudformation


pferrel:
What are you basing the doc change on?

markbar:
Experience and other user feedback I found online. When trying to deploy it rolls back and logs this error: "Value (us-east-1a) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-west-2b, us-west-2a, us-west-2c". Just updating the documentation to reflect the availabilityzones that are supported.

EmergentOrder:
This change is incorrect, as which availability zones will work varies based on user account & other factors.

PR #199, opened 2016-01-16
Fix test, fix travis yml, update dependencies, Fix SQL default issue, merge from pferrel v0.9.6

EmergentOrder:


pferrel:
getting a bit messy so lets merge and start using

Issue #203, opened 2016-02-01
Mixed usage of case class and class with serializable interface.


HyukjinKwon:
I see the mixed usage of `case` classes with serializable and normal classes with serializable.
Scala `case` class itself is serializable by default. I think it does not have to batter extending `Serializable` and also making them as normal classes for all `case` classes with serializable and normal classes with serializable.

For example,

There is a normal class at [RandomModel.scala#L3-L5
](https://github.com/PredictionIO/PredictionIO/blob/65a64c902ad7671dfb2fb0e29d8af806a01c55d0/examples/experimental/scala-local-friend-recommendation/src/main/scala/RandomModel.scala#L3-L5)

```scala
class RandomModel(
  val randomThreshold: Double
) extends Serializable
```

This can be just

```scala
case class RandomModel(
  val randomThreshold: Double
)
```

Another example is the class at [Engine.scala#L6-L9](https://github.com/PredictionIO/PredictionIO/blob/master/examples/experimental/scala-parallel-recommendation-custom-datasource/src/main/scala/Engine.scala#L6-L9).

```scala
case class Query(
  user: String,
  num: Int
) extends Serializable
```

`extends Serializable` can be removed as below:

```scala
case class Query(
  user: String,
  num: Int
)
```

Are they intendedly used in a mixed way?
I would like to submit a PR to simply fix them if this is really an issue.

yipjustin:
Yes. Please submit a PR. For these cases, a plain "case class" suffices.

HyukjinKwon:
Fixed in #205

PR #205, opened 2016-02-04
Mixed usage of case class and class with serializable interface.

HyukjinKwon:
#203

This PR fixes mixed usage of case class and class with serializable interface.

It simply corrects some obvious cases at examples but did not for core, tools and data.

As I am not aware of coding conversion for PredictionIO, I would appreciate if you look through this closely.

### DIFF
--- a/examples/experimental/scala-local-friend-recommendation/src/main/scala/FriendRecommendationPrediction.scala
+++ b/examples/experimental/scala-local-friend-recommendation/src/main/scala/FriendRecommendationPrediction.scala
@@ -1,7 +1,7 @@
 package io.prediction.examples.friendrecommendation
 
-case class FriendRecommendationPrediction (
-  confidence: Double,
+class FriendRecommendationPrediction (
+  val confidence: Double,
   // returning boolean acceptance to align with KDD 2012 scenario
-  acceptance: Boolean
-)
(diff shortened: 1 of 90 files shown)


yipjustin:
Hello Hyukjin,

Thanks for contributing. There are two key changes in this PR: 1. Converting a `case class Foo() extends Serializable` into `case class Foo()`.  2. Converting a `class Foo() extends Serializable` into a `case class Foo()`.

Case 1 is good, since it is redundant to mixin `Serializable` into a `case class`.

But we have to be careful with the case 2, since scala forbids case-to-case inheritance, in other words, we cannot have another `case class Bar() extends Foo()`. In these examples, users may want to extends an existing `TrainingData` and create multiple subclass, each serving slightly different purpose. Because of the case-to-case restriction, even if we have base `case class FOO()`, all subclasses still need to be a non-case-class, which defies this PR's purpose of "fixing mixed usage of case class and class with serializable interface."

HyukjinKwon:
@yipjustin I see. Thanks! Let me remove those changes.

HyukjinKwon:
@yipjustin Thanks for comments. I just removed the changes.

yipjustin:
LGTM

PR #210, opened 2016-02-26
Use https when submitting user email to pio

yevgenius:
When installing pio using the install.sh script, users that subscribe to
updates submit their email to a direct.prediction.io endpoint using
http. Email is PII and should be transmitted using https.


dszeto:
Thanks @yevgenius !

PR #211, opened 2016-03-02
Add TLS and simple key auth

yevgenius:
Add TLS and simple key auth to dashboard main page and engine server /stop and /reload endpoints


common/src/main/scala/io/prediction/authentication/KeyAuthentication.scala
```diff
@@ -0,0 +1,45 @@
...
+import spray.routing.{AuthenticationFailedRejection, RequestContext}
+import spray.routing.authentication._
+import spray.routing.directives.AuthMagnet
+import scala.concurrent.Future
+
+
+trait KeyAuthentication {
+
+  object ServerKey {
+    val serverConfig = ConfigFactory.parseFile(new File("conf/server.conf"))
```
dszeto: This expects the configuration file to live relative to where the `pio` command is run. How about using this method? (http://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigFactory.html#load-java.lang.String-)

common/src/main/scala/io/prediction/configuration/SSLConfiguration.scala
```diff
@@ -0,0 +1,52 @@
...
+import java.io.File
+import java.security.KeyStore
+import javax.net.ssl.{KeyManagerFactory, SSLContext, TrustManagerFactory}
+
+import com.typesafe.config.ConfigFactory
+import spray.io.ServerSSLEngineProvider
+
+trait SSLConfiguration {
+
+  private val serverConfig = ConfigFactory.parseFile(new File("conf/server.conf"))
```
dszeto: Same as above.

common/src/main/scala/io/prediction/authentication/KeyAuthentication.scala
yevgenius: Good point, thanks!

common/src/main/scala/io/prediction/configuration/SSLConfiguration.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  private val keyStoreResource = serverConfig.getString("io.prediction.server.ssl-keystore-resource")
+  private val password = serverConfig.getString("io.prediction.server.ssl-keystore-pass")
+  private val keyAlias = serverConfig.getString("io.prediction.server.ssl-key-alias")
+
+  private val keyStore = {
+
+    //Loading keystore from specified file
+    val clientStore = KeyStore.getInstance("JKS")
+    val inputStream = new FileInputStream(keyStoreResource)
```
dszeto: Just tested the latest code. Sorry that I did not realize this earlier. This also requires a special treatment. Guess we need to do something like `getClass().getClassLoader().getResource(keyStoreResource).getFile())`.

yevgenius:
@dszeto everything works now :)

dszeto:
@yevgenius :+1: 
I bumped into JCE issue on a fresh Java 8 install. Probably a good idea that we add a note about it (http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html). Merging this. Thanks!

PR #242, opened 2016-06-27
Make SSL and authKey param authentication optional

chanleesiq:


pferrel:
@dszeto Is this all that is required for "opt-in" style SSL, meaning the default is no SSL, then you can config to use it? If so we can create a PR for the ActionML fork after this is merged.

dszeto:
Duplicate of #244 / PIO-1.

Discussion continued in https://issues.apache.org/jira/browse/PIO-1

PR #249, opened 2016-07-07
[PIO-2] Rename the packages

yinxusen:
Rename all packages name from `io.prediction` to `org.apache.predictionio`.

Changed:

1. All packages names;
2. All imports related to those packages;
3. All code related to those packages such as `conf.set("io.prediction.xxx", xxx)`.
4. Others: configurations, bin.

Unchanged for now:

1. Files under docs/


yinxusen:
Let's wait to see whether it can pass tests, since it related to so many files.

Another thing I noticed: Code files under `examples/` use different  package names, e.g. `com.test`, `com.test1`, `org.template`. Shouldn't we change them into a unified package name?

dszeto:
@yinxusen Agree. In fact, some of these examples are probably outdated. Let's make a separate task to have those cleaned up.

dszeto:
LGTM. Merging.

Issue #243, opened 2016-06-30
https://docs.prediction.io/support/ down


indrekpihor:
Is this url coming back up anytime soon? Or will the SalesForce shut this site down for good?

franklinkim:
This seems to affect the whole `https` site thus making a lot of internal links on the site useless.
You can still reach it through `http` (i.e. http://docs.prediction.io/support/)

nqbao:
I hope this project won't become a deadend :(

ravishchawla:
Is this issue also effecting engines? I get an SSL Handshake exception when trying to build my engine.

dszeto:
Hi all, PredictionIO will continue as Apache PredictionIO (incubating). More details will follow this week. Sorry for the inconvenience meanwhile during the migration!

dszeto:
@ravishchawla The build process should succeed. The SSL exception comes from a second thread that should not have interfered with the main build process.

ravishchawla:
Yes, the build process did eventually complete, despite the error.

CodyErekson:
@dszeto: Any more details yet? I have a project that is on hold until I can get a new build up and running.

dszeto:
@CodyErekson Unfortunately I was not able to expedite the fix in infrastructure. I will instead drive towards a new patch and rely on Apache's infrastructure.

CodyErekson:
@dszeto: Thanks for the update and new information.

PR #250, opened 2016-07-14
[PIO-8] Remove upgrade check

Ziemin:
Apart from not being implemented, upgrade check was causing SSL Handshake exceptions during the engine building.

See also: #243


dszeto:
LGTM

PR #259, opened 2016-07-26
[PIO-14] Template gallery

Ziemin:
These changes contain:
- templates.yaml file containing template descriptions following the same schema. All templates from the old site have been exported here
- gallery_generator.rb with a function to generate mardkdown from the yaml file
- updated guidelines on how to submit a new template


dszeto:
This is great. Thanks!

Issue #251, opened 2016-07-15
Installation script doesn't work


ex0rcist:
Installation script `$ bash -c "$(curl -s https://install.prediction.io/install.sh)"`

does not work, I think mostly because it's trying to download `https://github.com/PredictionIO/PredictionIO/releases/download/v${PIO_VERSION}/${PIO_FILE}`, but `PredictionIO/PredictionIO` repo is unavailable any more.

AndersonSingh:
Hey, I encountered the same problem as well. I got everything working by building from the source files.

djones:
When this project was moved the releases and tags have been lost https://github.com/apache/incubator-predictionio/releases that is why the files in the install script are missing.

PR #263, opened 2016-08-02
Set correct Github URL for downloading releases

djones:
This will get us closer to solving #251 

The second part to solving this issue will involve restoring the release tags that have been lost since the project was moved.

Because version `0.9.6` is specified in the install script, only that tag is really needed for now. Once the tag has been pushed, Github should create a release and the file needed to make the install script work will be generated.

Would someone with appropriate permissions be able push the tag?

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -290,7 +290,7 @@ echo -e "\033[1;36mStarting PredictionIO setup in:\033[0m $pio_dir"
 cd ${TEMP_DIR}
 if [[ ! -e ${PIO_FILE} ]]; then
   echo "Downloading PredictionIO..."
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
Is this about https://issues.apache.org/jira/browse/PIO-10?

djones:
Ah yes, I didn't see the issues over at https://issues.apache.org

dszeto:
Hi @djones, thank you so much for the PR! Unfortunately, we cannot distribute binaries via GitHub. Even though we push the tags, GitHub will only create links for downloading sources. We are working on a new Apache branded release that can live on the ASF distribution network. Please stay tuned!

djones:
OK thanks @dszeto closing now.

PR #264, opened 2016-08-02
[PIO-12] Add a rule for checking Scala documentation style

HyukjinKwon:
This PR adds the rule for checking Scala documentation style.

This rule will check below:

- Wrong

```
/**
 *
...
```

- Correct

```
/**
  *
...
```

We can't just enable `org.scalastyle.scalariform.ScalaDocChecker` because it forces  all the classes, traits, methods, types and properties to have documentation which is overwhelming.

I ran the checking and then corrected the Java style documentation to Scala documentation according to http://docs.scala-lang.org/style/scaladoc.html.


HyukjinKwon:
@dszeto Could you please take a look?

dszeto:
@HyukjinKwon This is awesome. Thank you so much!

PR #267, opened 2016-08-05
[PIO-17] Testing infrastructure

Ziemin:
This PR adds basic integration tests written in python and also a new travis configuration using them. Travis will execute unit tests as before, as well as the new integration tests in 4 concurrent builds (3 builds for different environment configurations for integration tests).
Example for the current build is [here](https://travis-ci.org/Ziemin/incubator-predictionio/builds/150144156) .

What is more, there is an introduction of a docker image, which helps both in setting up travis tests and in running tests locally yourself. The image itself is currently pulled from my own repo on docker hub - ziemin/pio-testing, which should be moved to some official PredictionIO's repository in the future.

Given that there are no official engines, which could be downloaded and would work properly with the current develop version, integration tests have slightly modified version of recommendation engine in related directory.

More details can be found in /tests/README.md


dszeto:
I think a discussion just started on dev@ regarding the inclusion of ex-official templates in the main codebase. If that happens, these tests could use those templates, and in the other way, these templates would always be kept up to date.

dszeto:
@chanlee514 @Ziemin I notice there is a standalone Spark cluster in the Docker container but nowhere in the test seems to use it (I couldn't see any `pio` commands with `--master` argument). Is it planned in the future?

chanlee514:
Yes we'll add tests for standalone Spark in the future.

PR #277, opened 2016-08-17
specify the output file explicitely when installing via install.sh script

svalaskevicius:
the `-O` option on my system has downloaded the file to `develop` instead of the *.tar.gz file that is used later in the script


EmergentOrder:
Thanks, merged

svalaskevicius:
Np, btw the download link in this script on the master branch is broken and I don't know what it should be instead :) (github seems to have removed the packages download feature)

PR #279, opened 2016-08-19
[PIO-26] Integrate Apache RAT for license checking

chanlee514:


.travis.yml
```diff
@@ -6,6 +6,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
 
 branches:
   only:
```
smarthi: Don't think u need to include license header for travis.yaml; most other Apache Top projects don't do it. Ensure that u exclude this from rat-ccheck.

chanlee514: I saw that Spark did this (https://github.com/apache/spark/blob/master/.travis.yml) but if you think it'd be better to exclude this, I'll go ahead and update. Thanks!

smarthi: Ok, lets keep it then. ITs not necessary but no harm in keeping it.

dszeto:
This is great! Thanks @chanlee514 !

chanlee514:
I'm not sure if license check is the reason for the delay. As you can see here (https://travis-ci.org/apache/incubator-predictionio/builds/153982829) there's is little difference in execution time for unit testing. In my case running license check locally is almost instant.

PR #283, opened 2016-08-24
[PIO-28] Console refactor

Ziemin:
I moved some logic from functions in console to a separate package - commands. These function should now mostly return Expected[Result] or MaybeError, which are basically aliases for Either[String, Result]. With the introduced changed it will be easier to implement the Admin Server, which could simply make use of the same logic as Console.

In addition to that I removed template get and list commands as well as upgrade, because they are no longer going to be supported. Currently typing one of this commands will result with appropriate error.


dszeto:
This looks good to me, but due to the size of this PR, it would be nice to have extra pairs of eyes. Could @pferrel / @EmergentOrder help? This PR would be the first step to making admin API possible.

dszeto:
I am merging this. Thanks @Ziemin !

vlordier:
then most examples need to be updated to reflect this change, such as 
http://predictionio.incubator.apache.org/templates/similarproduct/quickstart/ etc

What replaces 
`pio template list / get ? ` (if anything) ?

PR #284, opened 2016-08-25
Update tapster.html.md

docherty:
Formatting: added a line break before a bullet list so markdown renders it correctly


dszeto:
Thanks @docherty !

PR #290, opened 2016-09-04
[PIO-37] mkdir: make parents if necessary

aniruddha-adhikary:
```
mkdir: cannot create directory ‘/xyz/PredictionIO/vendors’: No such file or directory
```


dszeto:
Great. Thanks @aniruddha-adhikary !

PR #297, opened 2016-09-20
Update document

djKooks:
While installing from source with Postgresql, there were no default JDBC driver file, so I add guide for download and setup for this. Please give a comment if there is any problem.


djKooks:
One more, while I am working on tutorial, there are some broken parts in document that are using angle bracket. It looks it is hard to control on markdown text.

http://stackoverflow.com/questions/20610244/how-to-use-angle-brackets-in-url-on-github-flavored-markdown

How about changing this to square bracket?
This is a suggestion, so please let me know if this is not a good choice.

dszeto:
Thanks for your contribution! I don't have a problem with replacing angle brackets with square brackets. The only ask from me is to keep it consistent throughout the whole documentation. Thanks!

djKooks:
@dszeto Okay thanks! I'll keep update it if there is no problem.

dszeto:
@djKooks No problem. Please keep updating. Thanks for your contribution!

djKooks:
@dszeto I reviewed on documents, but only the part that I fixed shows the problem, and other parts used angle bracket shows fine. You can find the problem in link below.

http://predictionio.incubator.apache.org/templates/recommendation/quickstart/

I'm not sure to remove all angle bracket because most of them are working fine, and if I start changing, lots of part need to be changed and maybe some one could be confused in it. Maybe it would be to let it in current state because it is pretty small bug. I'll think about other way to show it clear without replacing it.

PR #300, opened 2016-09-22
[PIO-35] Add integration tests for official templates

chanlee514:
**Changes:**

- Integration test fetches templates from Github, instead of storing local copy.
- Install git in Docker image.

**Notes**

- As can be seen in `tests.py`,  3 tests are run by default on Travis: *['BasicAppUsecases', 'EventserverTest', 'ScalaParallelRecommendationTest']*. One can change this using `TEST_NAMES` param in `.travis.yml` 
- Each scenario should be updated to use standalone Spark cluster for more stable behavior ([https://issues.apache.org/jira/browse/PIO-36](url)). I've done some work on this, and will update as [PIO-36].
- I've excluded 'JavaParallelEcommercerecommendationTest' for now since the template fails to load org.jblas dependency before test execution.

### DIFF
--- a/docs/manual/source/customize/dase.html.md.erb
+++ b/docs/manual/source/customize/dase.html.md.erb
@@ -1,3 +1,20 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
 ---
 title: Implementing DASE
 ---
--- a/docs/manual/source/datacollection/analytics-ipynb.html.md.erb
+++ b/docs/manual/source/datacollection/analytics-ipynb.html.md.erb
@@ -1,3 +1,20 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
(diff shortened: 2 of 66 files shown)


.travis.yml
```diff
@@ -68,7 +69,7 @@ env:
 
 before_install:
   - unset SBT_OPTS JVM_OPTS
-  - docker pull ziemin/pio-testing
+  - docker pull chanlee514/pio-testing
```
dszeto: I think we should put this under `predictionio` org on Docker Hub. We are missing something that automatically build and tag these testing container images properly. Let me work on it and we can modify our integration test to pick up the proper container.

tests/pio_tests/data/eventserver_test/partially_malformed_events.json
```diff
@@ -1,10 +1,10 @@
...
+  {
     "event" : "test",
     "entityType" : "test",
     "entityId" : "t2"
   },
   {
-    "event" : "malformed-event" 
+    "event" : "malformed-event"
   }
 ]
```
dszeto: Would be nice to have a newline here.

tests/pio_tests/scenarios/basic_app_usecases.py
```diff
@@ -61,17 +61,13 @@ class BasicAppUsecases(BaseTestCase):
   def setUp(self):
     random.seed(3)
     self.log.info("Setting up the engine")
-
-    template_path = pjoin(
-        self.test_context.engine_directory, "recommendation-engine")
+    template_path = "https://github.com/chanlee514/template-scala-parallel-recommendation"
```
dszeto: Any changes to your fork's template? We should use official ones.

.travis.yml
chanlee514: That sounds good. Once this is done, I think we should add documentation on how to experiment/contribute to the Docker image as well.

tests/pio_tests/scenarios/basic_app_usecases.py
chanlee514: They're temporary forks with org namespace updates (io.prediction >> org.apache.predictionio) to make the tests work. Maybe I can create a PR to each of the official template repos following v0.10 release, and then update this?

dszeto: That's a good idea. In fact, the templates have been donated and are in the process of migrating to Apache. I will ping you when that's done, so you can start using those new Git URLs as well.

PR #314, opened 2016-10-17
Update target version on install script

djKooks:
I updated version name on install script. It cannot find file path while installing.


djKooks:
@dszeto Got it! Thanks~

PR #325, opened 2016-11-08
Add git to Dockerfile

chanlee514:
For fetching templates via github

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -30,7 +30,8 @@ RUN apt-get install -y \
     python3-pip \
     postgresql postgresql-contrib \
     openjdk-8-jdk \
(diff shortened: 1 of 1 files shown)


pferrel:
Do we have Alpine as the container OS? This seems to be the standard Docker is pushing but this looks like Debian. 

BTW You can merge this yourself now. We, as a group, like to review on Github but after a fair amount of time (3-4 days) without response feel free to ping someone or just merge.

chanlee514:
The tests use latest version of standard Debian Linux image on Docker. I think it may be a good idea to use Alpine though since it's much smaller in size and may speed up the tests. 

And thanks, I will go ahead and merge this now.

PR #328, opened 2016-11-30
[PIO-47] [PIO-51] Stateless build & pio commands outside engine directory

chanlee514:
Changes:
* manifest.json is removed completely. `pio build`is stateless, and no longer writes anything to metadata storage.
* engineFactory (ex. "org.template.recommendation.RecommendationEngine") is used as **engineId** instead of a randomly-generated string


dszeto:
@chanlee514 Can you add a JIRA ticket for this and describe the motivation behind it on that ticket please? After that please edit the PR's title to have the ticket number.

dszeto:
Please try to run `pio train` / `pio deploy` commands outside of an engine directory and see if it returns helpful error messages.

chanlee514:
Right now, the commands throw java.io.FileNotFoundException regarding engine.json. May be better if we throw an error message saying the commands should be executed inside engine directory.

core/src/main/twirl/org/apache/predictionio/workflow/index.scala.html
```diff
@@ -65,13 +63,6 @@
         <tr><th>Average Serving Time</th><td>@{f"${avgServingSec}%.4f"} seconds</td></tr>
         <tr><th>Last Serving Time</th><td>@{f"${lastServingSec}%.4f"} seconds</td></tr>
         <tr><th>Engine Factory Class (Scala/Java)</th><td>@{engineInstance.engineFactory}</td></tr>
-        <tr>
```
chanlee514: Maybe we can search for the relevant .jar files in engine target/ directory?

chanlee514:
@pferrel: I agree, and that would be possible with a general metadata registry server which would be the next step. The reason I started by removing manifest.json is because the previous database-dependent build was a roadblock for PredictionIO in PaaS platforms such as Heroku. 

Information in engine.json is written during `pio train` as engineInstance.

dszeto:
@pferrel For `pio build` we probably can enhance it such that it can take a working directory for compilation to run. If we look at some other build tools, they also rely on having certain files in the current working directory.

For `pio train` and `pio deploy`, I absolutely agree that it should be able to run from anywhere as the most advanced scenario. My personal take to this change would be something comparable to a local HBase and distributed HBase setup. It would be ideal if someone can start playing with PredictionIO without setting up any external services (not even a SQL database). I believe Chan's change here is a partial step to make the learning curve less steep, and path ways for us to untangle engine registration so that it could become a feature that can be more easily documented and understandable.

Do you currently have any production deployments that rely heavily on engine IDs and versions? That would be a bigger immediate concern regarding this change.

pferrel:
ok, this looks like a good step

@dszeto we would not do `pio build` we would do `sbt build` in the directory that has the template and this is fine IMO. The confusion users have is over pio commands.

BTW @chanlee514 @dszeto Are we thinking of a new command, something like `pio register` that would add metadata to the metastore? This would need to be run every time the engine.json changed for instance? It would also do not compile? Is there an alternative? What state does this leave us in? After the push, what action create binary (I assume `pio build`) what action adds metadata to the metastore (I assume `pio build`) So this is only about changing the instance id and getting rid of manifest, right?

If so I'm totally cool with this, a good step I agree. I guess I'm using this PR as a forum for discussion and will move it to @dev

pferrel:
One more question. After push how do we run the PredictionServer or train on multiple machines? In the past this required copying the manifest.json and making sure binaries are in the same location on all machines.

pferrel:
@dszeto asked: "Do you currently have any production deployments that rely heavily on engine IDs and versions? That would be a bigger immediate concern regarding this change."

Yes. That is why so many questions above. We create a pio driver machine that runs training and is then taken down. This is in sync with bringing up Spark and taking it down. The pio driver machine is really part of the cluster and for the UR (and other PIO templates we have) the predict part runs without Spark. So this requires we use some engine instance id across machines to find data in the metastore. We also have deployments with 2 EventServers and 2 PredictionServers behind a load balancer. This requires that the same metadata is available to several different machines running, what I've come to see as 3 independent processes, ES, PS, train. They are independent except for the metadata.

dszeto:
@chanlee514 @pferrel I have moved the discussion to dev@.

pferrel:
too much noise in @dev, I have a simple question. 

If I train on one machine, how do I deploy on another. Currently I copy the manifest--well actually the entire directory--from the train machine to the deploy machine. 

This is actually a very important use case because the train machine has the Spark driver on it so it may have to be very large (and temporary, but the deploy machine can be smaller and far cheaper. Not to mention having multiple PredictionServers behind a load balancer.

pferrel:
@chanlee514 said: 

If the absolute filepath is the same in the machines (since SHA-1 hash of
filepath is used as engineVersion), there would be no changes in the above
workflow. You can just copy the engine directory and deploy. If not, you
can also provide engine-id, engine-version as CLI arguments.

There's some additional ongoing work to make the process more robust and
clearer, such as enabling engine-directory as CLI input, help messages, and
adding tests to cover such scenarios. I don't think the PR will be merged
before then.

pferrel:
The file path is required to be the same on any 2 machines and is explicitly in the metadata. If you have the deploy machine dir in a different path than the train machine dir (actually the build dir in 0.10.0) you'll get the completely incomprehensible "one for one" failure. At least this is one reason for it. 

This IMO is  a major confusion point in PIO. Therefore I still think the ID should be explicit and engine-id and engine-version should be eradicated completely. 

Can someone please tell me why we need anything more than 

 - a dataset-id  (app)
 - a model-id, which points to metadata used to create a model from a dataset
 - the metadata would contain the specific engine.json data used and the dataset-id (actually the appName in the engine but we need to remove this too) so it would connect the model-id to the dataset-id with all data needed to create the model from the dataset.

In all commands (but one) this would allow the metadata to be retrieved from the metastore.

We can then re-do the CLI to be completely stateless, and not require anything be copied between machines except a common knowledge of IDs.

I will write up a proposal for this, maybe it will make sense but at least we can discuss it in concrete terms.

jimlyndon:
+1 

Actually ran into this issue a couple of months ago. Luckily myself and our dev ops guy immediately realized the issue (abs paths required) and fixed our process: We build on the driver machine, bundle the entire template dir, deploy on several web tiers that match the driver's paths exactly (all behind a load balancer).

However this is not very obvious. Especially for users coming from a ruby, node, etc., background who want to launch this in heroku, etc., and are not use to the "strict" ways of java land.

pferrel:
Actually the manifest may be gone but the requirement of the same absolute path is maintained, I believe. A hash of the path to the template code is used to create the id and is part of the metadata for the engine. So the "same path" remains but the manifest goes away. Correct me @chanlee514 if I'm wrong.

We are trying to create a roadmap that makes these hidden dependencies more clear and removing those that we can.

chanlee514:
@pferrel: Yes the hash of engine directory path is used as engineVersion (which kind of acts as the id). 

However, I'm working on making updates to this PR so that users can provide the directory path as CLI argument for `pio build/train/deploy/undeploy` commands. This will enable users to run pio commands anywhere, and hopefully make this issue clearer.

pferrel:
Cool, that's awesome. This news means the scope has changed, which is great but brings up the visibility of the id. 

Could I suggest that the user-visible engine-instance-id be settable by the user as a string. The path to template code is already in the metadata--great. Hash the path by default and report it or allow it optionally to be set with a string.

As you know I'm proposing that our goal be to have one PIO service that uses dataset and engine ids to send input and make queries like a DB with several human named tables. Allowing the user to set the ids makes it possible to have human readable ones used directly in this future REST API. Like 

 - `GET /engines/amazon/_query {json query body}` for a query to the "Amazon" engine
 - `GET /engines/netflix/_query {json query body}` for a query to the "Netflix" engine

Substitute any id you want and use with any template you want. BTW this method is used in several different REST APIs from MongoDB to Elasticsearch where tables are named by humans and the ids used in REST.

PR #343, opened 2017-02-01
new engine for linear regression

mgcdanny:
This is an example of a linear regression using the BFGS algorithm


dszeto:
Thanks @mgcdanny !

PR #345, opened 2017-02-12
[PIO-30] Set up a cross build for Spark 2.0 and Scala 2.11

chanlee514:
Built on top of #295.
For more details, please refer to  [https://github.com/apache/incubator-predictionio/pull/295](url).

**Important notes**

By default, `make-distribution.sh` will build for *Scala 2.10 & Spark 1.6*. This is to ensure that the change does not break current templates when building and locating engine/dependency jar files. In order to build for *Scala 2.11 & Spark 2.0*, run `make-distribution.sh -Dbuild.profile=scala-2.11`.

There needs to be some additional updates before this change is officially released:

* Install scripts and instructions
* `build.sbt` in all official templates (to get the proper Spark version depending on the user-provided build profile)
* Dockerhub test image ([https://hub.docker.com/r/predictionio/pio-testing](url))


shimamoto:
Hi @chanlee514 

I am interested in Spark 2.x support. But this PR bothers me a little. Do you mean that the true nature of this PR is that we can choose either Spark 1.6 or Spark 2.x when to run `make-distribution.sh`? I mean we cannot choose which Scala version to use. If we choose 1.6, Scala version is determined 2.10 because Spark 1.x is built with Scala 2.10 by default, otherwise 2.11.

But #295 comment says that we can configure the following:
```
-Dbuild.profile=scala-2.11 -Dspark.version=1.6.0
```
If we configure this, PredictionIO should download the Spark source package and build with Scala 2.11 support(http://spark.apache.org/docs/1.6.3/building-spark.html#building-for-scala-211).

Is this the same as what you imagined? Could you inform me of the future policy of Spark and Scala support?

chanlee514:
Hi @shimamoto, thanks for pointing that out. I'll make changes to enable `-Dbuild.profile=scala-2.11 -Dspark.version=1.6.0`, which is a simple matter of handling the build args in `build.sbt`.

Also, I think it may be a good idea to rename the build profiles to **spark1.6** and **spark2.0**. The main intention of the cross-build is to maintain backward compatibility when upgrading to spark 2.0 (which has many advantages, but requires scala version >= 2.11).

And lastly, `install.sh`, as well as the README files, are not yet updated to reflect the new changes, but I will work on that soon.

shimamoto:
@chanlee514 Thanks to take a look my comment!

We work on storage support for Elasticsearch 5.x with another PR(#352). This essentially does not require a separate build while allowing flexibility to use a different backend support.

Perhaps we could propose a different way, which not -D parameters, depending on the future policy of Spark and Scala support. Thus, could you inform me of the policy if it works for you?

dszeto:
@shimamoto There isn't really a policy right now but we should be on the right path if we eventually add support of different combinations like you said (Spark 2.x + Scala 2.10). What would be your suggestion based on this?

shimamoto:
Hi @dszeto 

My suggestion is based on #352. This is what I see moving forward.

- Define the versions of Scala to build against in the `crossScalaVersions` setting.
- I’d be happy if it would be possible to deprecate Spark 1. We keep using the class (especially SQLContext) that keeping for backward compatibility for a while.
- Or the code that depends on version of Spark makes it abstract, and separate build. The jar that depends on Spark 1 is in lib/spark and the other is in lib/extra. In order to use Spark 2, one would replaced like ES5.

Note that this is just my imagination for now. Thus, I would like to try this out if this suggestion takes a turn for the better. For that, I think I need to some more time. Do you have any thoughts?

shimamoto:
Just for your information.
I tried my hand at removing unnecessary dependences and setting the `crossScalaVersions` as well.
https://github.com/shimamoto/incubator-predictionio/commit/ad0cace0507a951a160e29b3e1e7d9deb58aba22

At the tools project, type `+assembly`. Then, create all versions jars.

This is just temporary. I still have things to do. Sooner or later, I will open a PR.

PR #347, opened 2017-02-14
Unit is not Unit value

takezoe:
`Unit` is not Unit value. `()` is an only value of Unit type in Scala.

```
scala> Unit
res1: Unit.type = object scala.Unit

scala> () == Unit
<console>:12: warning: comparing values of types Unit and Unit.type using `==' will always yield false
       () == Unit
          ^
res2: Boolean = false
```

### DIFF
--- a/core/src/main/scala/org/apache/predictionio/controller/Engine.scala
+++ b/core/src/main/scala/org/apache/predictionio/controller/Engine.scala
@@ -208,7 +208,7 @@ class Engine[TD, EI, PD, Q, P, A](
       Doer(algorithmClassMap(algoName), algoParams)
     }
 
-    val models = if (persistedModels.exists(m => m.isInstanceOf[Unit.type])) {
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/apache/predictionio/controller/Engine.scala
```diff
@@ -222,7 +222,7 @@ class Engine[TD, EI, PD, Q, P, A](
 
       val models = algorithms.zip(persistedModels).map { case (algo, m) =>
         m match {
-          case Unit => algo.trainBase(sc, pd)
+          case () => algo.trainBase(sc, pd)
```
takezoe: In particular, maybe this case clause doesn't match.

marevol:
@dszeto Could you check and merge this pull request?
He is my co-worker and Scala expert.

dszeto:
Thanks for your reminder @marevol . I was planning to do so after fixing integration tests.

PR #353, opened 2017-02-26
Update link to working template of OpenNLP-SentimentAnalysis-Template

infoquestsolutions:
Updated template OpenNLP-SentimentAnalysis-Template to be compatible with Apache PIO.


dszeto:
@vshwnth2 Are you still maintaining your template? Are you okay with this change? Will merge this if there isn't a reply in a couple days.

PR #357, opened 2017-03-09
[PIO-51] Properly locate engine directory

chanlee514:
When --engine-dir is passed as CLI argument, use this instead of
user.dir

### DIFF
--- a/tools/src/main/scala/org/apache/predictionio/tools/console/Console.scala
+++ b/tools/src/main/scala/org/apache/predictionio/tools/console/Console.scala
@@ -734,7 +734,7 @@ object Console extends Logging {
     }
   }
 
-  def getEngineInfo(jsonFile: File): EngineInfo = {
+  def getEngineInfo(jsonFile: File, engineDir: String): EngineInfo = {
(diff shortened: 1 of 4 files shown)


tools/src/main/scala/org/apache/predictionio/tools/console/Console.scala
```diff
@@ -734,7 +734,7 @@ object Console extends Logging {
     }
   }
 
-  def getEngineInfo(jsonFile: File): EngineInfo = {
+  def getEngineInfo(jsonFile: File, engineDir: Option[String]): EngineInfo = {
```
dszeto: @chanlee514 should we just pass in a `String` here? It doesn't look like there would be a case for `None`. It also looks like `o.a.p.tools.Common.getEngineDirPath` has already taken care of getting the current working directory, so I think it should be safe to just drop using `user.dir` completely.

chanlee514: @dszeto Right I'll make the changes. Thanks!

PR #358, opened 2017-03-09
Update event serialization for ES5

marevol:
- Modify Event index mapping (change properties field to keyword type)
- Use UTC time format for eventTime/creatationTime


dszeto:
Thanks!

PR #361, opened 2017-03-13
Add solution for Hbase failure after disk full

linamy85:
Due to some issues of ZooKeeper, It takes some effort to have Hbase recovered from failure caused by full disk.


dszeto:
Awesome. Thanks @linamy85 !

PR #364, opened 2017-03-17
[PIO-30] Spark 2 support on sbt way

shimamoto:
The goal of this PR is:
- Define the versions of Scala to build against in the crossScalaVersions setting
    - Scala 2.10 and 2.11
- At the project, run `make-distribution.sh`. Then, create all versions gzs
    - PredictionIO_2.10-0.11.0-$VERSION.tar.gz <- This is intended for Spark 1 users
    - PredictionIO_2.11-0.11.0-$VERSION.tar.gz <- This is intended for Spark 2 users
- SQLContext maintain the status quo
    - I wish support for Scala 2.10 is deprecated as of next release
- Versions of Hadoop are 2.6+, which has something to do with storage hdfs
    - 2.6 is in lib/spark(default)
    - 2.7 is in lib/extra

@chanlee514 @dszeto
I did #345.

## TODO
Currently, only the traditional version of the test works. It needs to be added new version (Spark 2) tests.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,12 @@ javacOptions in (ThisBuild, compile) ++= Seq("-source", "1.7", "-target", "1.7",
 
 json4sVersion in ThisBuild := "3.2.10"
 
-sparkVersion in ThisBuild := "2.1.0"
+sparkVersion in ThisBuild := {
+  scalaBinaryVersion.value match {
+    case "2.10" => "1.6.3"
+    case _ => "2.1.0"
+  }
+}
 
 val pioBuildInfoSettings = buildInfoSettings ++ Seq(
   sourceGenerators in Compile <+= buildInfo,
--- a/build.sbt
+++ b/build.sbt
@@ -15,15 +15,15 @@
  * limitations under the License.
  */
 
-import UnidocKeys._
-
 name := "apache-predictionio-parent"
 
-version in ThisBuild := "0.11.0-SNAPSHOT"
+version in ThisBuild := "0.11.0-v1-SNAPSHOT"
 
 organization in ThisBuild := "org.apache.predictionio"
 
 scalaVersion in ThisBuild := "2.10.5"
+
+crossScalaVersions in ThisBuild := Seq(scalaVersion.value, "2.11.8")
 
 scalacOptions in ThisBuild ++= Seq("-deprecation", "-unchecked", "-feature")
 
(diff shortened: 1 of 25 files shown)


dszeto:
@shimamoto I realized the original build system was quite inflexible, so I went ahead and did this in the last couple days: https://travis-ci.org/apache/incubator-predictionio/builds/212014483. It's very close to working with tests across different dependencies. My apologies without syncing up with you.

The build system on that branch basically allows you to do this:
```
./make-distribution.sh -Dscala.version=2.10.6 -Dspark.version=2.1.0 -Dhadoop.version=2.7.3 -Delasticsearch.version=5.2.2
```
The script does not cross build, but `crossScalaVersions` in `build.sbt` is already defined, so it's possible to cross build and cross publish artifacts. `make-distribution.sh` could possibly be extended to produce more than one tarballs as well.

Also, before `docker-compose` becomes available on Apache Jenkins, I am using that as a multi-build just to make sure things compile across versions: https://builds.apache.org/job/incubator-predictionio-multibuild/

Could you take a look at this branch and see how it looks to you? It would be great to converge our changes.

shimamoto:
@chanlee514 I looked over the branch :)
I would like to confirm couple of things.

- Which versions are you going to make prebuilt packages?
- Do you mean that predictionio also supports versions that require users to build Spark(ex. Scala 2.10 & Spark 2.10)? This is just my personal opinion but, .....my impression is that 2.10 is intended for Spark 1 users and 2.11 is intended for Spark 2 users.
- ES5 has `elasticsearch-spark-13` dependency. We need to use `elasticsearch-spark-20` for Spark 2.

dszeto:
@shimamoto We can't really release binary packages on Apache yet until we have reviewed all packaged transitive dependencies comply with Apache policies. It will be a non-trivial exercise. Many people want to ship 0.11 as soon as possible, so it would probably be the next release. We will still require users to build their own package, so cross building everything in one shot would take too much time for them.

The goal is mostly about being flexible to use different dependencies instead of enforcing a set. We probably will only officially test Scala 2.10 + Spark 1.6, and Scala 2.11 + Spark 2. I tried other combos but haven't been able to make them work yet.

Agree the ES dep need to be fixed.

shimamoto:
@dszeto I understand your situation.
This PR is put EoM (Ease of Maintenance) before flexible because this leads to Easy of Extension in future. But as you say that the most important thing is flexible right now, this PR probably won't be able to achieve that goal. I think I will close this.

The goal of us is that users can use the standard sets without building by themselves (i.e. without run `make-distribution.sh`), which is preferable that the latest. Otherwise users need to build, and this was going to be the next step. But I think that's a very hard thing to do right now.

Please continue your good work.

dszeto:
I would still like to merge some of your changes if you are okay with that. Thanks!

PR #365, opened 2017-03-23
Remove duplicate settings.

shimamoto:
Thanks to merge some of #364 changes. However, since there are places where it looks some mistakes (duplication etc.), I fixed it.

### DIFF
--- a/tools/build.sbt
+++ b/tools/build.sbt
@@ -27,8 +27,6 @@ libraryDependencies ++= Seq(
   "io.spray"               %% "spray-testkit"  % "1.3.3" % "test",
   "org.specs2"             %% "specs2"         % "2.3.13" % "test")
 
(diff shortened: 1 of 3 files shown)


tools/build.sbt
```diff
@@ -27,8 +27,6 @@ libraryDependencies ++= Seq(
   "io.spray"               %% "spray-testkit"  % "1.3.3" % "test",
   "org.specs2"             %% "specs2"         % "2.3.13" % "test")
 
-dependencyOverrides += "org.slf4j" % "slf4j-log4j12" % "1.7.18"
-
```
shimamoto: Here is chosen this revision without this setting,isn't it?

dszeto: There was probably a SLF4J provider conflict before. Should be good to remove.

PR #367, opened 2017-04-06
[PIO-59] Use /dev/urandom to create access keys.

jsyrjala:
Creating access keys based on /dev/random may take a long time
on virtualized or clound environment for example more than
10 minutes.


dszeto:
I am convinced that this PR does not decrease the security of generating random app access keys. I will merge if there is no objection.

PR #368, opened 2017-04-10
Update linux distro url

steven-haddix:
The distro url appears to be moved to apache.org. Updating linux manual installation page.


dszeto:
Hey @steven-haddix , sorry for the late response and thanks for the PR. A couple comments about this PR.

The Linux manual installation page is supposed to be used with binary distributions of PredictionIO, which we are tackling here: https://issues.apache.org/jira/browse/PIO-69

Also Apache encourages the use of their distributions mirrors instead of hitting their archive server directly, thus this instruction: http://predictionio.incubator.apache.org/install/install-sourcecode/#downloading-source-code. It would be great to find a way for `wget` to work automatically with Apache mirrors.

PR #370, opened 2017-04-12
Add EVENTDATA_REFRESH for elasticsearch

marevol:
To improve indexing performance for event data, it's better to be able to set refresh parameter.
In PIO, refresh is true by default because updated data need to be made visible immediately.
To disable refresh on indexing actions, in this fix, set it as below:

    PIO_STORAGE_SOURCES_ELASTICSEARCH_EVENTDATA_REFRESH=false


dszeto:
Thanks @marevol ! Do you know if the REST client allows configuration by file? For the old ES1 client, one could supply an `elasticsearch.yml` file for the client to read in the rest of the configuration. I took a look at the REST client documentation and could not see something equivalent.

marevol:
AFAIK, Rest client does not have a configuration file.
I think it's a simple wrapper of HttpClient.

dszeto:
LGTM. Thanks!

PR #373, opened 2017-04-24
Improve insert performance for ES storage

marevol:
- Change properties and n to not indexed field
- Set refresh to false (_version does not need to be refreshed)

and modified pio version in tests to pass travis tests.


dszeto:
LGTM. By the way, are you working with ES 5.2/5.3? @mars bumped into issues with ES 5.1.

marevol:
Thank you for merging it.
Our production environment runs on ES 5.1, and my local dev is on ES 5.3.

PR #374, opened 2017-04-26
ES storage improvement/refactoring

marevol:
- Use the same ID generation for ES in event data
- Reuse RestClient instance in ESLEvents
- Replace 0 with false in mapping
- Set Content-Type as application/json


dszeto:
LGTM. I'd like to suggest us to get into the habit of filing JIRA tickets for all changes. It will make the release notes writer easier to figure out what's been changed.

PR #377, opened 2017-05-02
Update analytics-ipynb.html.md.erb

vaghawan:
While running the spark console in ipython, the current commands gives error, so to mitigate that I've added %pylab inline instead of pylab inline which seems to give the problem. This has been checked in my case.


dszeto:
Thanks!

vaghawan:
You welcome, well I would love to commit more, but I can't push anything in the livedoc. So it disappointed me.

dszeto:
To obtain committership, one would contribute and earn a nomination from a PMC member. Once the PMC voted favorably, the nominee will gain committership (the ability to push code directly to Apache repo). You are highly encouraged to make further contributions if committership is your goal. Please see http://apache.org/dev/. All Apache projects are governed similarly according to rules laid by the Apache Software Foundation.

vaghawan:
Ok, thanks, I would certainly love to contribute.

PR #379, opened 2017-05-03
Fix docs: pio template is no longer supported.

takezoe:


hyesun03:
I'm having trouble using prediction io.
Can you tell me what command can be use instead of `pio template`?

takezoe:
@hyesun03 Use `git clone` instead as:
```
git clone https://github.com/apache/incubator-predictionio-template-similar-product.git tapster-episode-similar
```

takezoe:
Merged to `livedoc` branch.

PR #383, opened 2017-05-11
[PIO-65] Cache downloaded jars in Travis build

takezoe:
Sometimes Travis build fails in downloading jars. Maybe caching them makes Travis build stable and also decreases build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,11 @@ services:
 
 sudo: required
 
-cache: false
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot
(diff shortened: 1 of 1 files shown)


takezoe:
Once close and reopen to re-run Travis build.

dszeto:
We should perhaps ping them again so we don't have to be stuck with Travis: https://issues.apache.org/jira/browse/INFRA-13543

takezoe:
Certainly! However this fix would improve current situation if docker-compose will not be available on ASF Jenkins soon.

dszeto:
Merged. Thanks @takezoe !

PR #385, opened 2017-05-26
[PIO-72] Fix class loading for pio-shell

chanlee514:


bin/compute-classpath.sh
```diff
@@ -17,36 +17,35 @@
...
-ASSEMBLY_DIR="${FWDIR}/assembly"
-
 if [ -n "$JAVA_HOME" ]; then
   JAR_CMD="$JAVA_HOME/bin/jar"
 else
   JAR_CMD="jar"
 fi
 
 # Use pio-assembly JAR from either RELEASE or assembly directory
+ASSEMBLY_DIR="${FWDIR}/assembly"
```
dszeto: I think we have a bug here and in our testing Docker builds. If you try this on a completely clean code tree, do `./make-distribution.sh` and directly run `bin/pio`, it stops working.

PR #389, opened 2017-06-05
[PIO-91] Fixed hadoop-hdfs artifact missing error

shimamoto:
I made a mistake when I reviewed dependencies.
Basically, the problem is due to unavailability of the hadoop-hdfs jars.


takezoe:
LGTM

PR #393, opened 2017-06-16
[PIO-94] Fix to show stacktrace for errors thrown in `queries.json` REST API

mars:
We were getting intractable errors from `queries.json` requests, like this one without a stacktrace:

```
[ERROR] [ServerActor] Query '{
  "user": "000",
  "item": "000"
}' is invalid. Reason: Expected object but got JNothing
```

This pull request adds stacktraces to these errors using the pattern already present elsewhere in `CreateServer.scala`.

[JIRA issue](https://issues.apache.org/jira/browse/PIO-94)


dszeto:
LGTM. Thanks @mars !

PR #391, opened 2017-06-13
[PIO-92] Bump HBase to 1.2.6 (and fix deprecation warnings)

takezoe:

PR #395, opened 2017-06-21
[PIO-92] Bump HBase to 1.2.6

takezoe:
This request upgrade only HBase dependency. Some methods have been deprecated, but they are not fixed in this request for backward compatibility.


takezoe:
I found HBase version inconsistency in documentation and build configuration: https://github.com/apache/incubator-predictionio/pull/391#issuecomment-310151519
So I close this request and confirm the correct state at first.

PR #401, opened 2017-06-28
[PIO-72] Fix class loading for pio-shell

chanlee514:


mars:
Back in May, we fixed an intermittent class loading problem by [making a change to stabilize the classpath](https://github.com/mars/incubator-predictionio/commit/9ecc77628aba347454073e9919096a8fc8e0b952) in our fork of incubator-predictionio. Would you be open to adding that to this PR as well?

chanlee514:
@mars Certainly. I've already merged it for my branch [here](https://github.com/chanlee514/incubator-predictionio/commits/pr-372). I still haven't had time to look into the shade rule error for Heroku UR. I'll update the PR once this has been resolved.

dszeto:
LGTM. Thanks @chanlee514 @mars !

BrianOn99:
I am using predicitonio 0.11 with postgres, and meet this issue `java.lang.ClassNotFoundException: jdbc.StorageClient` when using `pio-shell` when trying the tutorial https://predictionio.incubator.apache.org/datacollection/eventmodel/

As 0.12 is not released, I tried to apply this patch to 0.11 by appending `pio-data-jdbc-assembly-0.11.0-incubating.jar` to the `--jars` flag in `pio-shell` file, but then another error pops up
```
java.sql.SQLException: No suitable driver found for jdbc:postgresql://localhost/pio
        at java.sql.DriverManager.getConnection(DriverManager.java:689)
        at java.sql.DriverManager.getConnection(DriverManager.java:247)
```

Is it necessary to patch `Common.scala` and recompile?

Or is there another way to do the same thing as `PEventStore.aggregateProperties` as shown in the tutorial without `pio-shell`?

I really hope to have some help to be shown in the tutorial otherwise other newcomers like me will be confused.

mars:
Hi @BrianOn99,

Adding that `--jars` option to `pio-shell` command is the right solution, and then the "No suitable driver found" error can be solved by adding the Postgres driver to your PredictionIO install:

1. download [Postgres JDBC driver](https://jdbc.postgresql.org/download.html) (probably the newest one for Java 8)
2. put it in the PredictionIO distribution's `lib/` directory (this directory is sibling to the `bin/` directory where the `pio` command is located; any jars in that directory are automatically added to the classpath for `pio` commands)

We're working on releasing 0.12!

BrianOn99:
@mars Thanks for the help.  Yes the jdbc driver is not loaded correctly, though I have the jar downloaded.  I need to add `export CLASSPATH` like this in pio-shell
```
if [[ "$1" == "--with-spark" ]]
then
  echo "Starting the PIO shell with the Apache Spark Shell."
  # Get paths of assembly jars to pass to spark-shell
  . ${PIO_HOME}/bin/compute-classpath.sh
  shift
  export CLASSPATH
  ${SPARK_HOME}/bin/spark-shell --jars ${ASSEMBLY_JAR},/PredictionIO-0.11.0-incubating/lib/spark/pio-data-jdbc-assembly-0.11.0-incubating.jar $@
else
```
and then every thing works.

I note that `export CLASSPATH` exist in `bin/pio-class`, so in my environment `pio status`, `pio train`, etc, can work, but `pio-shell` does not call `bin/pio-class` so `CLASSPATH` is not exported.  Is this fixed upstream or my setting is wrong?

BrianOn99:
Good news to hear that 0.12 will be released soon!

mars:
Yes @BrianOn99, I do believe [class loading for pio-shell is fixed](https://github.com/apache/incubator-predictionio/blob/develop/bin/pio-shell#L59) for the next release, or if you `make-distribution.sh` on main `develop` branch, you'll get these fixes now.

PR #404, opened 2017-07-07
[PIO-100] Remove SLF4J warning message on pio command

marevol:


dszeto:
LGTM. I checked the dependency tree and looks like this is the easiest way out. Many transitive dependencies rely on `slf4j-log4j12`.

PR #409, opened 2017-07-14
Fix use of case class in recommendation examples (docs)

takezoe:
According to this template fix: https://github.com/apache/incubator-predictionio-template-recommender/pull/18


shimamoto:
LGTM!

PR #410, opened 2017-07-14
Fix use of case class in classification examples

takezoe:
According to this template fix: https://github.com/apache/incubator-predictionio-template-attribute-based-classifier/pull/11


shimamoto:
LGTM!

PR #412, opened 2017-07-15
[PIO-105] Batch Predictions

mars:
JIRA issue [PIO-105](https://issues.apache.org/jira/browse/PIO-105)

Provides a new `pio batchpredict` command using the same core logic & APIs as `pio deploy`. The only intentional difference between `batchpredict` and `deploy` is that the new `batchpredict` processes all queries via RDD#map.

As a result, a new engine constraint is added only for `batchpredict`:  
**All algorithm classes must be serializable.**

Reads from multi-object JSON input file. Example:

```json
{"user":"1"}
{"user":"2"}
{"user":"3"}
{"user":"4"}
{"user":"5"}
```

Writes to multi-object JSON output file (actually Hadoop partition files). Example:

```json
{"query":{"user":"1"},"prediction":{"itemScores":[{"item":"1","score":33},{"item":"2","score":32}]}}
{"query":{"user":"2"},"prediction":{"itemScores":[{"item":"5","score":55},{"item":"3","score":28}]}}
{"query":{"user":"3"},"prediction":{"itemScores":[{"item":"2","score":16},{"item":"3","score":12}]}}
{"query":{"user":"4"},"prediction":{"itemScores":[{"item":"3","score":19},{"item":"1","score":18}]}}
{"query":{"user":"5"},"prediction":{"itemScores":[{"item":"1","score":24},{"item":"4","score":14}]}}
```

See the included [console usage help](https://github.com/apache/incubator-predictionio/pull/412/files#diff-2cf174557564e09d52157be8e839fecf)

Thanks to @dszeto for architectural guidance and debugging assistance.

### DIFF
--- a/tools/src/main/scala/org/apache/predictionio/tools/commands/Engine.scala
+++ b/tools/src/main/scala/org/apache/predictionio/tools/commands/Engine.scala
@@ -263,7 +263,7 @@ object Engine extends EitherLogging {
     }
   }
 
-  /** Batch predict with an engine. 
+  /** Batch predict with an engine.
     *
     * @param ea An instance of [[EngineArgs]]
     * @param engineInstanceId An instance of [[engineInstanceId]]
--- a/tools/src/main/scala/org/apache/predictionio/tools/commands/Engine.scala
+++ b/tools/src/main/scala/org/apache/predictionio/tools/commands/Engine.scala
@@ -21,8 +21,9 @@ import org.apache.predictionio.core.BuildInfo
 import org.apache.predictionio.controller.Utils
 import org.apache.predictionio.data.storage
 import org.apache.predictionio.tools.EitherLogging
-import org.apache.predictionio.tools.{RunWorkflow, RunServer}
-import org.apache.predictionio.tools.{DeployArgs, WorkflowArgs, SparkArgs, ServerArgs}
+import org.apache.predictionio.tools.{RunWorkflow, RunServer, RunBatchPredict}
+import org.apache.predictionio.tools.{
+  DeployArgs, WorkflowArgs, SparkArgs, ServerArgs, BatchPredictArgs}
 import org.apache.predictionio.tools.console.Console
 import org.apache.predictionio.tools.Common._
 import org.apache.predictionio.tools.ReturnTypes._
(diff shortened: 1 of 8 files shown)


tools/src/main/scala/org/apache/predictionio/tools/commands/Engine.scala
```diff
@@ -262,6 +263,56 @@ object Engine extends EitherLogging {
...
+    batchPredictArgs: BatchPredictArgs,
+    sparkArgs: SparkArgs,
+    pioHome: String,
+    verbose: Boolean = false): Expected[(Process, () => Unit)] = {
+
+    val engineDirPath = getEngineDirPath(ea.engineDir)
+    val verifyResult = Template.verifyTemplateMinVersion(
+      new File(engineDirPath, "template.json"))
+    if (verifyResult.isLeft) {
+      return Left(verifyResult.left.get)
```
takezoe: Generally, avoiding return is preferred in Scala if it's possible. I think following is better:

```scala
verifyResult match {
  case x @ Left(err) => x
  case Right(_) => {
  	...
  }
}
```

or this is bit shorter:

```scala
verifyResult.right.flatMap { _ =>
  ...
}
```

However, there are some other `return`s in this file, so it might be better to fully refactor in an another pull request.

mars:
@takezoe thank you for the feedback. As a relatively-new Scala programmer I really appreciate this kind of review.

I am a bit hesitant to make these changes. I'm trying to maintain likeness with the [`CreateServer.scala`](https://github.com/mars/incubator-predictionio/blob/e7c6ebd8cfe2d4a150319025876520fc39be9a34/core/src/main/scala/org/apache/predictionio/workflow/CreateServer.scala) code, to minimize differences in prediction behavior between `pio deploy` and `pio batchpredict`. Any of these stylistic points should probably be matched in CreateServer, so that it continues to be easy to reason about their similarity.

takezoe:
@mars These changes should work as same as original code completely, but difference of code appearance might cause confusion. So these kind of changes might have to be done at once (in an another pull request).

I think it's a reasonable decision that even if you don't apply these changes in this pull request. 👍

dszeto:
LGTM. Great work @mars !

Agree with @takezoe 's comments. I think there are a few follow up tasks we should do.

1. Refactor common code that exists in both `CreateServer` and `BatchPredict`.
2. Brush up Scala coding style as pointed out by @takezoe .
3. Document the new feature in the doc site.

dszeto:
Created https://issues.apache.org/jira/browse/PIO-110 and https://issues.apache.org/jira/browse/PIO-111 as follow-ups. Thanks @mars for the feature and @takezoe for the feedback!

PR #417, opened 2017-08-01
[PIO-66] JIRA and release process for PredictionIO

shimamoto:
https://issues.apache.org/jira/browse/PIO-66?page=com.atlassian.jira.plugin.system.issuetabpanels:all-tabpanel

### DIFF
--- a/docs/manual/data/nav/main.yml
+++ b/docs/manual/data/nav/main.yml
@@ -319,6 +319,9 @@ root:
     url: '#'
     children:
       -
+        body: 'Release Cadence'
+        url: '/resources/release/'
+      -
         body: 'Developing Engines with IntelliJ IDEA'
         url: '/resources/intellij/'
       -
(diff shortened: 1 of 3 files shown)


dszeto:
Hey, what do you think about putting Release Cadence under the Getting Involved section? We should probably clean up the Resources section because it sounds pretty generic.

takezoe:
I think adding a new section like "For Developers" is good for Release Cadence and other documentation for committers. Further, we should clean up the whole index in the left sidebar. Contents are good, but the index is not much kind. I think we can classify them more intelligible.

Anyway, we should do it on an another ticket. How about merging this addition as is and creating a new ticket about re-structuring the documentation?

shimamoto:
Currently there isn't section to put documents for committers. It is better to clean up not only Resources section but also the others. It might be a good idea to do it in another task.

dszeto:
@shimamoto @takezoe Yes that sounds good. I'll merge this. Thanks!

PR #419, opened 2017-08-08
Update index.html.md.erb

Sequential-circuits:
Since Scala 2.10 is used to compile projects and that version is critical seems to me it should be included as a requirement. My 2 Cents.


takezoe:
Scala is not prerequisite because you don't have to install Scala by hand. If you just run `make-distribution.sh`, PredictionIO is compiled with the default version of Scala (currently 2.10.6) automatically.

PR #420, opened 2017-08-11
[PIO-106] Elasticsearch 5.x StorageClient should reuse RestClient

mars:
Implements [PIO-106](https://issues.apache.org/jira/browse/PIO-106)

This PR moves to a singleton Elasticsearch RestClient which has built-in HTTP keep-alive and TCP connection pooling. Running on this branch, we've seen a 2x speed-up in predictions from the Universal Recommender with ES5, and the feared "cannot assign requested address" 😱 Elasticsearch connection errors have completely disappeared. Running `pio batchpredict` for 160K queries results in only 7 total TCP connections to Elasticsearch. Previously that would escalate to ~25,000 connections before denying further connections.

**This fundamentally changes the interface for the new [Elasticsearch 5.x REST client](https://github.com/apache/incubator-predictionio/tree/develop/storage/elasticsearch/src/main/scala/org/apache/predictionio/data/storage/elasticsearch)** introduced with PredictionIO 0.11.0-incubating. With this changeset, the `client` is a single instance of [`org.elasticsearch.client.RestClient`](https://github.com/elastic/elasticsearch/blob/master/client/rest/src/main/java/org/elasticsearch/client/RestClient.java).

🚨 **As a result of this change, any engine templates that directly use the Elasticsearch 5 StorageClient would require an update for compatibility.** The change is this:

### Original 

```scala
val client: StorageClient = … // code to instantiate client
val restClient: RestClient = client.open()
try {
  restClient.performRequest(…)
} finally {
  restClient.close()
}
```

### With this PR

```scala
val client: RestClient = … // code to instantiate client
client.performRequest(…)
```

*No more balancing `open` & `close` as this is handled by using a new `CleanupFunctions` hook added to the framework in this PR.*

[Universal Recommender](https://github.com/actionml/universal-recommender) is the only template that I know of which directly uses the ES StorageClient outside of PredictionIO core. See example [UR changes for compatibility with this PR](https://github.com/heroku/predictionio-engine-ur/compare/esclient-singleton).

### Elasticsearch StorageClient changes

* reimplemented as singleton
* installs a cleanup function

See [StorageClient](https://github.com/apache/incubator-predictionio/compare/develop...mars:esclient-singleton?expand=1#diff-2926f4cfd93ccb02320e2a9503ccd223)

### Core changes

[...]

PR #372, opened 2017-04-20
Elasticsearch basic HTTP authentication

mars:
Add optional username-password configuration for the new Elasticsearch 5 client; in `pio-env.sh` config:

```bash
# Optional basic HTTP auth
PIO_STORAGE_SOURCES_ELASTICSEARCH_USERNAME=my-name
PIO_STORAGE_SOURCES_ELASTICSEARCH_PASSWORD=my-secret
```

These credentials are sent in each Elasticsearch request as an HTTP Basic Authorization header.

Enables use of public-cloud, hosted Elasticsearch clusters, such as [Bonsai on Heroku](https://elements.heroku.com/addons/bonsai).

I'm looking into adding test coverage. (I have the Docker test suite setup now.)

PR #421, opened 2017-08-14
Elasticsearch 5.x singleton client with authentication

mars:
Fixes both [PIO-106](https://issues.apache.org/jira/browse/PIO-106) & [PIO-114](https://issues.apache.org/jira/browse/PIO-114), replacing https://github.com/apache/incubator-predictionio/pull/372 & https://github.com/apache/incubator-predictionio/pull/420. These are combined because they each heavily revise the same class.

## Authentication

Add optional username-password configuration for the new Elasticsearch 5 client; in `pio-env.sh` config:

```bash
# Optional basic HTTP auth
PIO_STORAGE_SOURCES_ELASTICSEARCH_USERNAME=my-name
PIO_STORAGE_SOURCES_ELASTICSEARCH_PASSWORD=my-secret
```

These credentials are sent in each Elasticsearch request as an HTTP Basic Authorization header.

Enables use of public-cloud, hosted Elasticsearch clusters, such as [Bonsai on Heroku](https://elements.heroku.com/addons/bonsai).

## Singleton client

This PR moves to a singleton Elasticsearch RestClient which has built-in HTTP keep-alive and TCP connection pooling. Running on this branch, we've seen a 2x speed-up in predictions from the Universal Recommender with ES5, and the feared "cannot assign requested address" 😱 Elasticsearch connection errors have completely disappeared. Running `pio batchpredict` for 160K queries results in only 7 total TCP connections to Elasticsearch. Previously that would escalate to ~25,000 connections before denying further connections.

**This fundamentally changes the interface for the new [Elasticsearch 5.x REST client](https://github.com/apache/incubator-predictionio/tree/develop/storage/elasticsearch/src/main/scala/org/apache/predictionio/data/storage/elasticsearch)** introduced with PredictionIO 0.11.0-incubating. With this changeset, the `client` is a single instance of [`org.elasticsearch.client.RestClient`](https://github.com/elastic/elasticsearch/blob/master/client/rest/src/main/java/org/elasticsearch/client/RestClient.java).

🚨 **As a result of this change, any engine templates that directly use the Elasticsearch 5 StorageClient would require an update for compatibility.** The change is this:

### Original 

```scala
val client: StorageClient = … // code to instantiate client
val restClient: RestClient = client.open()
try {
  restClient.performRequest(…)
} finally {
  restClient.close()
}
```

### With this PR

```scala
val client: RestClient = … // code to instantiate client
client.performRequest(…)
```

*No more balancing `open` & `close` as this is handled by using a new `CleanupFunctions` hook added to the framework in this PR.*

[Universal Recommender](https://github.com/actionml/universal-recommender) is the only template that I know of which directly uses the ES StorageClient outside of PredictionIO core. See example [UR changes for compatibility with this PR](https://github.com/heroku/predictionio-engine-ur/pull/7/files).

### Elasticsearch StorageClient changes

* reimplemented as singleton
* installs a cleanup function

See [StorageClient](https://github.com/apache/incubator-predictionio/compare/develop...mars:esclient-singleton-with-auth?expand=1#diff-2926f4cfd93ccb02320e2a9503ccd223)

### Core changes

A new [`CleanupFunctions`](https://github.com/apache/incubator-predictionio/compare/develop...mars:esclient-singleton-with-auth?expand=1#diff-2a958821ac58f019fbce38540c775f19) hook has been added which enables developers of storage modules to register anonymous functions with `CleanupFunctions.add { … }` to be executed after Spark-related commands/workflows. The hook is called in a `finally { CleanupFunctions.run() }` from within:

* `pio import`
* `pio export`
* `pio train`
* `pio batchpredict`

Apologies for the huge indentation shifts from the requisite try-finally blocks:

```scala
try {
  // Freshly indented code.
} finally {
  CleanupFunctions.run()
}
```


storage/elasticsearch/src/main/scala/org/apache/predictionio/data/storage/elasticsearch/StorageClient.scala
```diff
@@ -18,27 +18,84 @@
...
+import org.apache.predictionio.workflow.CleanupFunctions
 import org.elasticsearch.client.RestClient
+import org.elasticsearch.client.RestClientBuilder.HttpClientConfigCallback
 
 import grizzled.slf4j.Logging
 
-case class ESClient(hosts: Seq[HttpHost]) {
-  def open(): RestClient = {
+object ESClient extends Logging {
+  var _sharedRestClient: Option[RestClient] = None
```
takezoe: Is this client accessed from multi threads? If yes, I think adding `volatile` modifier to this field is necessary and `open()` and `close()` should be synchronized.

storage/elasticsearch/src/main/scala/org/apache/predictionio/data/storage/elasticsearch/StorageClient.scala
```diff
@@ -18,27 +18,84 @@
...
       case e: Throwable =>
         throw new StorageClientException(e.getMessage, e)
     }
   }
+
+  def close(): Unit = {
+    if (!_sharedRestClient.isEmpty) {
+      _sharedRestClient.get.close()
+      _sharedRestClient = None
+    }
```
takezoe: Typically, `Option.get()` is hated in the Scala world. You can write as follows instead:
```scala
_sharedRestClient.foreach { client =>
   client.close()
  _sharedRestClient = None
}
```

storage/elasticsearch/src/main/scala/org/apache/predictionio/data/storage/elasticsearch/ESPEvents.scala
```diff
@@ -107,37 +107,32 @@ class ESPEvents(client: ESClient, config: StorageClientConfig, index: String)
...
+                ("eventId" -> eventId)))
+          val entity = new NStringEntity(compact(render(json)), ContentType.APPLICATION_JSON)
+          val response = client.performRequest(
+            "POST",
+            s"/$index/$estype/_delete_by_query",
+            Map("refresh" -> ESUtils.getEventDataRefresh(config)).asJava)
+          val jsonResponse = parse(EntityUtils.toString(response.getEntity))
+          val result = (jsonResponse \ "result").extract[String]
+          result match {
+            case "deleted" => true
```
takezoe: `true` at here is no effect because the return type us `Unit`. I know it came from original code, but we should remove it.

storage/elasticsearch/src/main/scala/org/apache/predictionio/data/storage/elasticsearch/ESPEvents.scala
```diff
@@ -107,37 +107,32 @@ class ESPEvents(client: ESClient, config: StorageClientConfig, index: String)
...
+            case _ =>
+              logger.error(s"[$result] Failed to update $index/$estype:$eventId")
               false
```
takezoe: Same as above.

storage/elasticsearch/src/main/scala/org/apache/predictionio/data/storage/elasticsearch/ESPEvents.scala
```diff
@@ -107,37 +107,32 @@ class ESPEvents(client: ESClient, config: StorageClientConfig, index: String)
...
           }
+        } catch {
+          case e: IOException =>
+            logger.error(s"Failed to update $index/$estype:$eventId", e)
+            false
```
takezoe: Same as above.

storage/elasticsearch/src/main/scala/org/apache/predictionio/data/storage/elasticsearch/StorageClient.scala
mars: Thanks for the hints here. I suspected this would be an issue. I'm investigating how to make this threadsafe.

mars:
Cheers @takezoe I addressed all your Scala style & usage suggestions. Still need to take care of the threadsafety issue with the singleton client.

takezoe:
> I addressed all your Scala style & usage suggestions.

👍 

> Still need to take care of the threadsafety issue with the singleton client.

Yes, I'm reviewing the PIO codebase from this point of view now, but still my understanding about the whole PIO is not enough...

mars:
Based on these [Scala Concurrency/Thread Safety docs](https://twitter.github.io/scala_school/concurrency.html#danger), I believe simply annotating `@volatile` will cause the synchronization needed for thread-safety in this case.

So, I updated this PR with that change.

mars:
🏁

takezoe:
@mars LGTM! but sorry for conflict with my previous commit: 6789dbeb71b1cc7f13a385032da6fbc3b8cf7a12

mars:
I will resolve these conflicts today and then merge this PR.

PR #424, opened 2017-08-22
[PIO-115] Implement Storage app & channel name-to-ID cache

mars:
Fixes [PIO-115](https://issues.apache.org/jira/browse/PIO-115)

When stress testing the Universal Recommender with high-concurrency HTTP/REST queries, we observed that Elasticsearch traffic was majority composed of requests resolving the Storage app's name & channel, over and over and over again! In this case, [each per-query call to `LEventStore.findByEntity`](https://github.com/heroku/predictionio-engine-ur/blob/master/src/main/scala/URAlgorithm.scala#L694) re-resolves the app name to an ID.

This changeset implements memoization for the function that performs these name-to-ID lookups, so that only one set of lookups is performed per process for each app+channel combination. As a result, we've seen overall throughput increase 📈 and error rate drop dramatically 📉.

This common optimization effects all storage backends, not just Elasticsearch.

### DIFF
--- a/data/src/main/scala/org/apache/predictionio/data/store/Common.scala
+++ b/data/src/main/scala/org/apache/predictionio/data/store/Common.scala
@@ -19,6 +19,7 @@
 package org.apache.predictionio.data.store
 
 import org.apache.predictionio.data.storage.Storage
+import scala.collection.mutable
 import grizzled.slf4j.Logger
 
 private[predictionio] object Common {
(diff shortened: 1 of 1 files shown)


data/src/main/scala/org/apache/predictionio/data/store/Common.scala
```diff
@@ -19,35 +19,46 @@
...
+              logger.error(s"Invalid channel name ${ch}.")
+              throw new IllegalArgumentException(s"Invalid channel name ${ch}.")
+            }
+          }
+
+          appNameToIdCache((appName, channelName)) = (app.id, channelId)
+          (app.id, channelId)
+        }.getOrElse {
+          logger.error(s"Invalid app name ${appName}")
+          throw new IllegalArgumentException(s"Invalid app name ${appName}")
```
dszeto: It looks like this whole block can be simplified to

```scala
appNameToIdCache.getOrElseUpdate((appName, channelName), {
  val appOpt = ...
  appOpt map { app =>
    ...
    val channelId: Option[Int] = ...
    // do not need to set appNameToIdCache here manually
    (app.id, channelId)
  } getOrElse {
    ...
  }
})
```

mars:
Thanks for your feedback @dszeto. I've addressed the code style & JIRA issue.

PR #426, opened 2017-08-29
Fix use of case class in textclassifier example

takezoe:
The text-classifier example is fixed in https://github.com/apache/incubator-predictionio-template-text-classifier/pull/14.


shimamoto:
This fix is LGTM!
But I wonder what this menu is... 🤔

PR #427, opened 2017-08-29
[PIO-116] PySpark Support

marevol:
This PR provides PySpark support with minimum PIO changes.

1. Support pyspark on pio-shell
2. Add python files to use pyspark
3. Add --main-py-file option to "pio train" to submit .py file to spark

Note that this provides only fixes for Spark 2.x.
(because this fixes expect to use SparkML)

Sample project is:
https://github.com/jpioug/predictionio-template-iris
(For prediction API, Scala code is used.)

### DIFF
--- a/bin/pio-shell
+++ b/bin/pio-shell
@@ -59,6 +59,15 @@ then
   . ${PIO_HOME}/bin/compute-classpath.sh
   shift
   ${SPARK_HOME}/bin/spark-shell --jars ${ASSEMBLY_JARS} $@
+elif [[ "$1" == "--with-pyspark" ]]
+then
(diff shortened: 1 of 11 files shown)


marevol:
Thank you for checking it. Replaced with if-else.

dszeto:
@marevol Looks like Python 3 is also a hard requirement? This does not work when I run `pyspark` with Python 2.

marevol:
Added __init__.py. It will work on Python 2.7.

dszeto:
Verified working with Python 2. Thanks!

marevol:
Thanks! I'll merge this PR tomorrow.

PR #431, opened 2017-09-06
Update templates.yaml - user name change

thomasste:


dszeto:
Verified that GitHub is forwarding repos from the previous user to the new user. Merging. Thanks @thomasste !

PR #432, opened 2017-09-07
[PIO-120] Process hangs if Elasticsearch is not available during train

mars:
Fixes [PIO-120](https://issues.apache.org/jira/browse/PIO-120)

This changeset ensures that the process exits gracefully after ES connection error.

Apologies for the indentation shift. This changeset simply wraps the body of the main function with:

```scala
try {
  // much existing code
} finally {
  CleanupFunctions.run()
}
```


mars:
Would be great to have this included in 0.12.0 release.

PR #433, opened 2017-09-09
Update webhooks.html.md.erb

aayush142128:
Webhook is provided by the application that provides data, so PredictionIO is responsible for exposing webhook connectors.


dszeto:
Merging. Thanks @aayush142128 !

PR #436, opened 2017-09-21
[PIO-125] Add Spark 2.2 support

marevol:


dszeto:
Hey @marevol , any idea with the failing test? I'm wondering if the new Guava dependency is not working fine with older Hadoop/Spark.

marevol:
For test failures on Travis, I think it's Travis problem. [My Travis](https://travis-ci.org/jpioug/incubator-predictionio/builds/282978334) result is passed.

Spark 2.2 seems to break Guava dependency...
@shimamoto looked into this problem.

marevol:
Re-run Travis.

shimamoto:
It is not supposed to need the new Guava dependency because Spark has Guava dependency.
But if this dependency isn't included, compile error occurs at data project.

When I investigated the cause, I discovered Spark 2.1.1 and Spark 2.2.0 were different Guava dependencies.

- Spark 2.1.1 : com.google.guava:guava:14.0.1
- Spark 2.2.0 : com.google.guava:guava:11.0.2

It is expected guava:14.0.1 in Spark 2.2 intrinsically.

The root cause is probably this:
https://github.com/sbt/sbt/issues/2861

The Apache Curator version referenced in Spark has changed in 2.2.0.
This has been thought to have some effect.

https://github.com/apache/spark/blob/v2.1.1/pom.xml#L130
https://github.com/apache/spark/blob/v2.2.0/pom.xml#L126

https://github.com/apache/curator/blob/2.4.0/pom.xml#L307
https://github.com/apache/curator/blob/apache-curator-2.6.0/pom.xml#L424

marevol:
Oops, my travis result missed some results...
I'll fix this PR. This problem might come from HBase.

marevol:
Re-run Travis.

marevol:
Spark 2.2 has guava dependency problem and [SPARK-21549](https://issues.apache.org/jira/browse/SPARK-21549).
I added a workaround for SPARK-21549.

marevol:
Re-run Travis.
I think it's better to reduce test matrix...

dszeto:
@marevol Apache installed `docker-compose` on their Jenkins finally. We can move these tests there for faster build time.

dszeto:
LGTM. Merging. Thanks @marevol !

PR #437, opened 2017-09-26
Update org.template references to org.example

sillystring13:
When eval is run according to instructions you get: 
Exception in thread "main" scala.reflect.internal.MissingRequirementError: object org.template.textclassification.AccuracyEvaluation not found.

It looks like the org.template references are stale after a refactor to org.example


takezoe:
LGTM!

PR #438, opened 2017-10-05
[PIO-129] Move CLI document

takezoe:
There are links to CLI document in the deploy section and collecting data section of the side menu, but if these links are clicked, the side menu is closed because these links have a hash like `/cli/#engine-commands`. I think that such unclear navigation would confuse readers.
https://predictionio.incubator.apache.org/cli/#engine-commands

I propose to remove these links from the deploy section and the collecting data section, and put a link to CLI document in the resource section without hash:

cli-doc1

In addition, put links to the CLI reference in documents of the event server and the engine server. A following screenshot shows a link to the event server CLI in the document about deployment:

cli-doc2

### DIFF
--- a/docs/manual/data/nav/main.yml
+++ b/docs/manual/data/nav/main.yml
@@ -59,9 +59,6 @@ root:
         body: 'Deploying as a Web Service'
         url: '/deploy/'
       -
-        body: 'Engine Command-line Interface'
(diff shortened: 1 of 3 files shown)


dszeto:
LGTM. Merging. Thanks @takezoe !

PR #439, opened 2017-10-05
[PIO-131] Fix Apache licensing issues for doc site

chanlee514:


docs/manual/source/install/install-sourcecode.html.md.erb
```diff
@@ -28,15 +28,20 @@ Download Apache PredictionIO (incubating) <%= data.versions.pio %> from an
 Apache [mirror](https://www.apache.org/dyn/closer.cgi/incubator/predictionio/<%=
 data.versions.pio %>/apache-predictionio-<%= data.versions.pio %>.tar.gz). 
 
+Verify this release using [signatures and checksums](https://www.apache.org/
+dist/incubator/predictionio/0.12.0-incubating/) and [project release KEYS](
```
dszeto: How about using `data.versions.pio` instead?

docs/manual/source/install/install-sourcecode.html.md.erb
```diff
@@ -28,15 +28,20 @@ Download Apache PredictionIO (incubating) <%= data.versions.pio %> from an
...
+https://www.apache.org/dist/incubator/predictionio/KEYS). Refer to
+[instructions](http://httpd.apache.org/download.cgi#verify) on how to verify.
```
dszeto: I think it would be better to copy those instructions over.

PR #447, opened 2017-11-17
Fix batchpredict for custom PersistentModel

mars:
Fixes [PIO-138](https://issues.apache.org/jira/browse/PIO-138)

Switches batch query processing from Spark RDD to a Scala parallel collection. As a result, the `pio batchpredict` command changes in the following ways:

* `--query-partitions` option is no longer available; parallelism is now managed by Scala's [parallel collections](http://docs.scala-lang.org/overviews/parallel-collections/overview.html)
* `--input` option is now read as a plain, local file
* `--output` option is now written as a plain, local file
* because the input & output files are no longer parallelized through Spark, memory limits may require that large batch jobs be split into multiple command runs.

This solves the root problem that certain custom PersistentModels, such as [ALS Recommendation template](https://github.com/apache/incubator-predictionio-template-recommender), may themselves [contain RDDs](https://github.com/apache/incubator-predictionio-template-recommender/blob/develop/src/main/scala/ALSModel.scala#L27), which cannot be nested inside the batch queries RDD. (See [SPARK-5063](https://issues.apache.org/jira/browse/SPARK-5063))

### DIFF
--- a/core/src/main/scala/org/apache/predictionio/workflow/BatchPredict.scala
+++ b/core/src/main/scala/org/apache/predictionio/workflow/BatchPredict.scala
@@ -18,7 +18,8 @@
 
 package org.apache.predictionio.workflow
 
-import java.io.Serializable
+import java.io.{BufferedWriter, File, FileWriter, Serializable}
+import java.net.URI
 
 import com.twitter.bijection.Injection
 import com.twitter.chill.{KryoBase, KryoInjection, ScalaKryoInstantiator}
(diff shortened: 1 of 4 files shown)


mars:
I'm currently testing this change with various engines and large batches.

mars:
Tested this new `pio batchpredict` with all three model types:

* ✅ custom PersistentModel (ALS Recommendation)
* ✅ built-in, default model serialization (Classification)
* ✅ null model (Universal Recommender)

This PR is ready to go!

mars:
BTW, I found performance for a large, 250K query batch is equivalent to the previous Spark RDD-based performance.

mars:
This PR stalled due to @dszeto's concerns about removing the distributed processing capability from `pio batchpredict`. I agree that distributed batch processing is optimal, but do not have a solution for the nested RDDs problem encountered for RDD-based persistent models.

PR #450, opened 2018-01-12
[PIO-179] bump up hbase client version and make it configurable

voran:
Since predictionio 0.12+ suggests hbase 1.2.6 by default, it makes sense to also use a the same version of the client instead of the now outdated 0.98.5. 

In the interest of flexibility, this change makes it configurable at build time by passing -Dhbase.version=...

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1131,7 +1131,7 @@ version.
 --------------------------------------------------------------------------------
 Binary distribution bundles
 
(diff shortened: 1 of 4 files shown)


dszeto:
@voran Thank you for your contribution! Would you please fix the license check test, file a ticket under Apache JIRA (https://issues.apache.org/jira/projects/PIO/issues) and change the subject of this PR to have the ticket number please?

dszeto:
LGTM. Thank you!

PR #455, opened 2018-06-12
fix "$pio eval" command

DanVisan:
"$pio eval" command parameters are broken: package name in command does not match the package name in Evaluation.scala

### DIFF
--- a/docs/manual/source/evaluation/metricbuild.html.md
+++ b/docs/manual/source/evaluation/metricbuild.html.md
@@ -67,7 +67,7 @@ object. We can run the following command to kick start the evaluation.
 ```
 $ pio build
 ...
-$ pio eval org.example.classification.AccuracyEvaluation org.template.classification.EngineParamsList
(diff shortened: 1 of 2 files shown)


dszeto:
@tom-chan I think @DanVisan is referring to https://github.com/apache/predictionio-template-attribute-based-classifier/blob/develop/src/main/scala/Evaluation.scala. That is what is being linked from this file right now.

tom-chan:
Got it. In that case do you mind updating other instances of `org.template.classification.` to `org.example.classification.` on this page, as well as docs/manual/source/evaluation/metricbuild.html.md? Thanks.

tom-chan:
Would be great if you also update `docs/manual/source/evaluation/metricbuild.html.md` in the same tutorial series. Thanks!

DanVisan:
@tom-chan i updated `docs/manual/source/evaluation/metricbuild.html.md`. Thanks for noticing!

tom-chan:
LGTM

PR #459, opened 2018-08-17
[PIO-158] Deprecate Scala 2.10 and Spark 1.x

shimamoto:


shimamoto:
@dszeto Could you take a look?

dszeto:
Looks like 2 major potentially breaking changes are in as well:

* Dropping Java 7 support
* Dropping Spark 1.3 to 1.5 support

Should we instead put out a deprecation notice for those as well before dropping?

shimamoto:
Well...I don't think we need to put out a deprecation notice because these versions no longer meet prerequisites we expose.
https://github.com/apache/predictionio/blob/livedoc/docs/manual/source/install/index.html.md.erb#L22

or rather, it's a bug and should be fixed. Would you like me to separate it from this PR?

dszeto:
I am merging this.

PR #462, opened 2018-09-05
[PIO-86] Docker support

marevol:
Please see [README](https://github.com/marevol/predictionio/blob/docker/docker/README.md).

### DIFF
--- a/docker/pio/Dockerfile
+++ b/docker/pio/Dockerfile
@@ -1,0 +1,71 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+FROM openjdk:8
+
+ENV PIO_TAG=v0.12.1
+ENV SCALA_VERSION=2.11.12
+ENV SPARK_VERSION=2.2.2
+ENV HADOOP_VERSION=2.7.7
+ENV ELASTICSEARCH_VERSION=5.5.3
+ENV PGSQL_VERSION=42.2.4
+ENV MYSQL_VERSION=8.0.12
+ENV PIO_HOME=/usr/share/predictionio
+
+RUN apt-get update && \
+    apt-get install -y dpkg-dev fakeroot && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt/src
+RUN git clone https://github.com/apache/predictionio.git
+WORKDIR /opt/src/predictionio
+RUN git checkout refs/tags/$PIO_TAG
+RUN bash ./make-distribution.sh \
+      -Dscala.version=$SCALA_VERSION \
+      -Dspark.version=$SPARK_VERSION \
+      -Dhadoop.version=$HADOOP_VERSION \
+      -Delasticsearch.version=$ELASTICSEARCH_VERSION \
+      --with-deb && \
+    dpkg -i ./assembly/target/predictionio_*.deb && \
+    rm -rf /opt/src/predictionio/*
+
+
+RUN cp /etc/predictionio/pio-env.sh /etc/predictionio/pio-env.sh.orig && \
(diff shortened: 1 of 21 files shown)


takezoe:
The official docker support is nice! and I was impressed your idea which composite PredictionIO's configuration by compositing docker-compose.yml by `-f` option!

However I think docker-compose is just for quick start for users, not for production use because we would need to create our own Spark cluster or Elasticsearch cluster if we need them outside of a PredictionIO instance. For this purpose, I think a fixed minimal structure of docker-compose.yml (e.g. PIO + PostgreSQL + Spark) is sufficient.

Further, if we offers the docker support officially, it would be nice if we can put a docker image to Apache registry (but I don't know how can we do it!):
https://hub.docker.com/r/apache/

marevol:
Thank you for your feedback.

> However I think docker-compose is just for quick start for users

I have a plan to support Kubernetes after this PR.
To do that, I consider that Kompose generates K8s' yml files for production use.
So, I think that this multiple docker compose files are for production.

takezoe:
Ah, your plan is that creating whole PredictionIO enviornment indcluding Spark cluster and storage cluster on the Kubernetes cluster, right? That was beyond my imagination, really exciting! If so, it would be nice if we can support HDFS, HBase and various managed services on the cloud as well.

I have some questions about Dockerfile, so I will make comments later. But basically agree with your Kubernetes support plan.

dszeto:
One way we can proceed is to mark the `Dockerfile` and the image it generates as official. Helper scripts and compose files can be marked as official reference. Good work @marevol !

docker/pio/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+WORKDIR /opt/src/predictionio
+RUN git checkout refs/tags/$PIO_TAG
+RUN bash ./make-distribution.sh \
+      -Dscala.version=$SCALA_VERSION \
+      -Dspark.version=$SPARK_VERSION \
+      -Dhadoop.version=$HADOOP_VERSION \
+      -Delasticsearch.version=$ELASTICSEARCH_VERSION \
+      --with-deb && \
+    dpkg -i ./assembly/target/predictionio_*.deb && \
+    rm -rf /opt/src/predictionio/*
```
takezoe: Building PredictionIO inside of the container is too slow. How about build outside of the container and just copy a binary to the container?

docker/pio/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+ENV MYSQL_VERSION=8.0.12
+ENV PIO_HOME=/usr/share/predictionio
+
+RUN apt-get update && \
+    apt-get install -y dpkg-dev fakeroot && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt/src
+RUN git clone https://github.com/apache/predictionio.git
```
takezoe: I think building the container from the current working copy would help development of PredictionIO as well. When we need to build a specific version of container, we can do it by checkout the tag.

docker/pio/Dockerfile
marevol: To create ivy cache, I put it at image build time.
`pio build` will take a long time(over 30min...) without the cache when running a container.
So, I do not think that it's acceptable to users...
Any other ideas are welcomed :)

docker/pio/Dockerfile
dszeto: Agree. I intend to use this as the base for CI, replacing the current test image.

docker/pio/Dockerfile
dszeto: It’s probably okay if we give the container more CPU and memory. The Ivy cache is a nice touch.

marevol:
> your plan is that creating whole PredictionIO enviornment indcluding Spark cluster and storage cluster on the Kubernetes cluster

Yes, and also you can use external clusters to specify it in docker-compose.[meta|event].yml.

marevol:
It's okay to merge this PR.
For a release procedure, I'll file a new issue on JIRA and update PMC.md.

marevol:
Updated:
- Set v0.13.0 as a default build tag
- Use --build-arg option to change a tag/branch

I'll merge this PR if there are no other concerns.

takezoe:
What is your opinion about this comment?
https://github.com/apache/predictionio/pull/462#discussion_r218608459

I have an another concern. Is it ok to include MySQL driver in the official docker image? I guess it's ok because this Dockerfile is in the official repository but docker image is delivered from DockerHub. So we can say: "It's not a distribution managed under ASF." But I'm not sure this logic is valid.

marevol:
> What is your opinion about this comment?

Does --build-arg option solve it?
You can change a branch or tag when building a docker image.

Just in case, I moved MySQL download process into pio_run.

takezoe:
> Does --build-arg option solve it?

Using `--build-arg` is nice! But I mean it would be nice if we can create a docker image on my working copy before commit and push something. But this is my opinion. Don't mind if you don't need it.

> Just in case, I moved MySQL download process into pio_run.

Good idea. This solution can be applied to setup of native distributions.

ahmettahasakar:
@marevol Is there any update on the kubernetes deploy documents? And is there any documentation about how to do it with multiple replicas?

Thank you

PR #465, opened 2018-09-20
Sync Template Gallery

takezoe:
Since the official templates now use `org.apache.predictionio` package they requires 0.11.0-incubating at least. I updated all `template.json` in the official templates with update for PredictionIO 0.13.0. Also the template gallery should be synchronized with them.


dszeto:
Thank you @takezoe !

PR #466, opened 2018-09-21
[PIO-168] ES 6.X support + patch version dependency updates

EmergentOrder:
This enables PIO to work with ES 6 and is backwards-compatible with earlier ES versions. 
The downside, and the reason this should go in a non-patch release is that ES 6 necessitates changes to the metadata structure, and this breaks existing PIO apps.
This is because ES 6 no longer allows creation of indices with multiple types (see https://www.elastic.co/blog/index-type-parent-child-join-now-future-in-elasticsearch )


dszeto:
Thanks @EmergentOrder ! Do you also have a migration tool for this change? It would be nice if existing users have a way to upgrade.

dszeto:
Please tag this PR with [PIO-168](https://issues.apache.org/jira/browse/PIO-168).

EmergentOrder:
Actually, according to the docs, existing indices, and so apps, should continue to work in 6.X (but will be disallowed in 7.X). I assume there is a documented ES migration process. There may be exceptions in template code, such as in the UR where the urindex is hot-swapped. I haven't tested this, but I will. 

Assuming that existing apps keep working, and on (re)creation they work, from the user's perspective I don't see a disruption / need for a PIO migration tool. This obviously doesn't hold for when we get to ES 7 though.

EmergentOrder:
Actually, scratch that, apps won't continue to work, although the indices will. We could add a fallback to the old indices in the code, but I think we should prefer to advise users to recreate and retrain their app when they upgrade.

dszeto:
I think we should provide a migration tool if possible. I will file another ticket for that.

dszeto:
@EmergentOrder is this ready to merge? I was waiting for build to pass but it looks like a transient error to me. Do you mind resolving the conflicts? It should trigger a new build.

EmergentOrder:
Fixed the conflicts. I agree, looked like a transient error.
Ready to merge, pending the next build.

EmergentOrder:
Not sure if that last failure was transient or what, but it looks like there are source dependencies on commons-lang3, so I added it to common/build.sbt.

takezoe:
In fact, Travis tests are sometimes unstable, so they fail randomly. It looks that adding commons-lang3 isn't related. Could you try again without it?

In my experience, if some of Travis tests fail, amend the latest commit and force push is an effective way to rerun Travis. :-P

takezoe:
👍

shimamoto:
@EmergentOrder Support for ES6 has bugs.
https://issues.apache.org/jira/browse/PIO-189

Is this something that you can do?

EmergentOrder:
@shimamoto Sure, I will take a look.

PR #467, opened 2018-09-21
Update release instructions for PMC

shimamoto:
@dszeto @chanlee514 I modified the release instructions slightly. Can you take a look?

Aside from this, I'm thinking about whether we should improve and automate the release process or not.

### DIFF
--- a/PMC.md
+++ b/PMC.md
@@ -115,7 +115,9 @@ for staging actual releases.
     * `svn delete 0.13.0`
 24. `svn commit -m "Apache PredictionIO 0.14.0"`
 25. Document breaking changes in https://predictionio.apache.org/resources/upgrade/.
-26. Send out an email to the following mailing lists: announce, user, dev.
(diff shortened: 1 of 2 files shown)


dszeto:
LGTM. Regarding automation, I think we can create scripts for some steps inside the doc.

dszeto:
@shimamoto If you don't mind, please add the missing step that makes a release on JIRA (https://issues.apache.org/jira/projects/PIO?selectedItem=com.atlassian.jira.jira-projects-plugin%3Arelease-page&status=no-filter).

shimamoto:
@dszeto Sure! You mean you want me to add the step that creates a next version after a release?

dszeto:
That would be great too. What I meant was we need to mark the version as released on JIRA. If you don’t see the option in JIRA please let me know.

shimamoto:
I fixed it!

dszeto:
LGTM. Merging. Thank you!

PR #469, opened 2018-09-21
[PIO-153] Allow use of GNU tar on non-GNU systems

dszeto:
@chanlee514 please review.

@shimamoto you may want to include this in your PMC doc update.

This addresses warnings when tarball is created using non-GNU tar (e.g. macOS). Here's an [example](https://superuser.com/questions/318809/linux-os-x-tar-incompatibility-tarballs-created-on-os-x-give-errors-when-unt).


shimamoto:
LGMT!

dszeto:
Thanks @shimamoto !

PR #470, opened 2018-09-22
[PIO-170] Upgrade sbt to 1.2.3

takezoe:
This upgrade requires dropping Scala 2.10 support (PIO-171). It's also required by replacing Spray with Akka-HTTP. We need to drop Scala 2.10 support first.


takezoe:
This upgrade requires dropping Scala 2.10 support. It's also required by replacing Spray with Akka-HTTP. We need to drop Scala 2.10 support first.

takezoe:
@shimamoto Can you take a look?

dszeto:
+1 from me

takezoe:
@dszeto Thanks!

build.sbt
```diff
@@ -310,10 +310,11 @@ parallelExecution in Global := false
 testOptions in Test += Tests.Argument("-oDF")
 
 printBuildInfo := {
+  val log = streams.value.log
   if (scalaBinaryVersion.value == "2.10")
-    streams.value.log.warn("Support for Scala 2.10 is deprecated. Please upgrade to a newer version of Scala.")
+    log.warn("Support for Scala 2.10 is deprecated. Please upgrade to a newer version of Scala.")
```
shimamoto: We should be able to delete this log because Scala 2.10 has been dropped.

build.sbt
```diff
@@ -310,10 +310,11 @@ parallelExecution in Global := false
...
   if (sparkBinaryVersion.value == "1.6")
-    streams.value.log.warn("Support for Spark 1.6 is deprecated. Please upgrade to a newer version of Spark.")
+    log.warn("Support for Spark 1.6 is deprecated. Please upgrade to a newer version of Spark.")
```
shimamoto: We should be able to delete this log because Spark 1.6 has been dropped.

shimamoto:
LGTM!

PR #471, opened 2018-09-22
[PIO-171] Drop Scala 2.10 and Spark 1.6 support

takezoe:
- [x] Build configuration update
- [x] Documentation update


build.sbt
```diff
@@ -17,24 +17,7 @@
 import PIOBuild._
 
 lazy val scalaSparkDepsVersion = Map(
```
dszeto: Maybe we can get rid of this `Map` as well?

docs/manual/source/install/index.html.md.erb
```diff
@@ -24,8 +24,8 @@ limitations under the License.
 It is **very important** to meet the minimum version of the following
 technologies that power Apache PredictionIO®.
 
-* Apache Hadoop 2.6.5 (optional, required only if YARN and HDFS are needed)
-* Apache Spark 1.6.3 for Hadoop 2.6
+* Apache Hadoop 2.7.3 (optional, required only if YARN and HDFS are needed)
+* Apache Spark 2.1.1 for Hadoop 2.7
```
dszeto: I wonder if we should put Hadoop 2.6.x/2.7.x, and Spark 2.0+ here. We still run CI on Spark 2.0, and the Hadoop requirement comes mostly from Spark.

build.sbt
takezoe: Yes, but we will add Scala 2.12 support very soon. So keeping this `Map` is better I think.

takezoe:
I added Spark 2.0.2 / Hadoop 2.6.5 test patterns and updated Spark version in `bin/pio-class.sh` and `tools/src/main/scala/org/apache/predictionio/tools/commands/Management.scala`.

dszeto:
@takezoe Looks great. I have no further comments. Thank you!

dszeto:
Just stumbled upon https://github.com/apache/predictionio/blob/develop/data/src/main/spark-1/org/apache/predictionio/data/SparkVersionDependent.scala. We probably should clean this up as well.

takezoe:
@dszeto Nice catch! I removed the file.

PR #475, opened 2018-10-02
[PIO-176] Hack in build.sbt for switching between Spark 1.x and 2.x s…

saurabh3091:
- Removed unmanagedSourceDirectories in data module
- refactored sqlSession code blocks to use SparkSession instead of SqlContext
- removed PPythonEventStore.scala as it was not being used anywhere.


saurabh3091:
Please review and let me know if I missed anything.
Thanks a lot for your support guys.

marevol:
PPythonEventStore is used in python/pypio.

shimamoto:
Yeah, it's necessary for python/pypio for now. We can probably remove it after supporting Spark ML natively.

PR #476, opened 2018-10-03
[PIO-176] Revert PPythonEventStore and add it to 'data' module

saurabh3091:


saurabh3091:
IntelliJ was unable to figure out the usage of PPythonEventOperator in python module.
How do you guys check this? (or do i need to find the word usage through whole project)

PR #478, opened 2018-10-10
[PIO-180] Trivial LiveDoc link change in readme

Wei-1:
[#docker-installation-for-predictionio](http://predictionio.apache.org/community/projects/#docker-installation-for-predictionio) change to [#docker-images](http://predictionio.apache.org/community/projects/#docker-images) in README.md


dszeto:
Should this be applied to `livedoc` as well?

Wei-1:
I think so. I will add this change in livedoc as well.

PR #482, opened 2018-10-15
[PIO-182] Add async methods to LEventStore

takezoe:
This pull request contains following updates:

- Add async (non-blocking) methods to `LEventStore` and `LJavaEventStore`
- Add Scaladoc about `scala.concurrent.ExecutionContext.Implicits.global` which is used in blocking methods

See also: https://lists.apache.org/thread.html/f14e4f8f29410e4585b3d8e9f646b88293a605f4716d3c4d60771854@%3Cuser.predictionio.apache.org%3E

### DIFF
--- a/data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
+++ b/data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
@@ -86,20 +86,18 @@ object LEventStore {
 
     import scala.concurrent.ExecutionContext.Implicits.global
 
-    val (appId, channelId) = Common.appNameToId(appName, channelName)
-
-    Await.result(eventsDb.futureFind(
-      appId = appId,
-      channelId = channelId,
-      startTime = startTime,
-      untilTime = untilTime,
-      entityType = Some(entityType),
-      entityId = Some(entityId),
+    Await.result(findByEntityAsync(
+      appName = appName,
+      entityType = entityType,
+      entityId = entityId,
+      channelName = channelName,
       eventNames = eventNames,
       targetEntityType = targetEntityType,
       targetEntityId = targetEntityId,
+      startTime = startTime,
+      untilTime = untilTime,
       limit = limit,
-      reversed = Some(latest)),
+      latest = latest),
       timeout)
   }
 
@@ -194,18 +192,16 @@ object LEventStore {
 
     import scala.concurrent.ExecutionContext.Implicits.global
 
-    val (appId, channelId) = Common.appNameToId(appName, channelName)
-
-    Await.result(eventsDb.futureFind(
-      appId = appId,
-      channelId = channelId,
-      startTime = startTime,
-      untilTime = untilTime,
+    Await.result(findAsync(
+      appName = appName,
       entityType = entityType,
       entityId = entityId,
+      channelName = channelName,
       eventNames = eventNames,
       targetEntityType = targetEntityType,
       targetEntityId = targetEntityId,
+      startTime = startTime,
+      untilTime = untilTime,
       limit = limit), timeout)
   }
 
--- a/data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
+++ b/data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
@@ -41,6 +41,9 @@ import scala.concurrent.duration.Duration
   *   -Dscala.concurrent.context.numThreads=1000 \
   *   -Dscala.concurrent.context.maxThreads=1000"
   * </pre>
+  *
+  * You can learn more about the global execution context in the Scala documentation:
+  * [[https://docs.scala-lang.org/overviews/core/futures.html#the-global-execution-context]]
   */
 object LEventStore {
 
--- a/data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
+++ b/data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
@@ -84,6 +84,7 @@ object LEventStore {
     latest: Boolean = true,
     timeout: Duration = defaultTimeout): Iterator[Event] = {
 
+    // Import here to ensure ExecutionContext.Implicits.global is used only in this method
     import scala.concurrent.ExecutionContext.Implicits.global
 
     Await.result(findByEntityAsync(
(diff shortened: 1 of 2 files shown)


takezoe:
@dszeto @shimamoto Could you take a look?

Chris who raised this issue wants to do adding the asynchrnous processing capability to other storages and engine server following this pull request.

data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
```diff
@@ -89,6 +103,57 @@ object LEventStore {
       timeout)
```
shimamoto: I think it is preferable to call the `findByEntityAsync` method from this method.

data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
```diff
@@ -142,4 +209,56 @@ object LEventStore {
       limit = limit), timeout)
```
shimamoto: same as above

data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
```diff
@@ -20,15 +20,27 @@ package org.apache.predictionio.data.store
...
+  * processors, parallelism is limited up to the number of processors.
+  *
+  * If this limitation become bottleneck of resource usage, you can increase the
+  * number of threads by declaring following VM options before calling "pio deploy":
+  *
+  * <pre>
+  * export JAVA_OPTS="$JAVA_OPTS \
+  *   -Dscala.concurrent.context.numThreads=1000 \
+  *   -Dscala.concurrent.context.maxThreads=1000"
+  * </pre>
```
takezoe: @dszeto Is this comment understandable? Since I'm not confident that it explains correctly and clearly, I'd like to request your confirmation.

data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
takezoe: Fixed.

data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
takezoe: Fixed as well.

data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
```diff
@@ -72,9 +84,61 @@ object LEventStore {
     latest: Boolean = true,
     timeout: Duration = defaultTimeout): Iterator[Event] = {
 
+    import scala.concurrent.ExecutionContext.Implicits.global
```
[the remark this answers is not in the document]
takezoe: Since using `ExecutionContext.Implicits.global` is bad habit in Scala basically, I wanted to make sure to use it only in these methods.

And your interpretation is correct. Even if we call it twice, it returns a same instance.

longliveenduro:
Any chance that this gets merged in develop soon? Would need to base my async work on that

takezoe:
@longliveenduro How do you feel about Scaladoc of `LEventStore.scala` in this pull request? Is explanation is enough and clear? I'm waiting for someone's review for that.

data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
takezoe: @Wei-1 Thanks for your feedback! I added line comment on importing as below:
```scala
// Import here to ensure ExecutionContext.Implicits.global is used only this method
import scala.concurrent.ExecutionContext.Implicits.global
```

Wei-1: btw, only `in` this method?

longliveenduro:
@takezoe The scaladoc is clear I think. 

But after some days of letting this settle and also reading this excellent blog: https://www.beyondthelines.net/computing/scala-future-and-execution-context/

I am not sure if it is really a good idea to use the "standard" Scala Execution Context for the "old blocking" code. Another idea would we (like proposed in the Blog) use a separate Thread pool, make the defaults the same as now (or maybe raise the allowed thread count to bypass/hack the problem from now on a bit for usages of the old code) and make the size of the thread pool configurable like the standard scala thread pool.

For example this blog suggests that a FixedThreadPool for blocking purposes might be a good idea:
https://www.cakesolutions.net/teamblogs/demystifying-the-blocking-construct-in-scala-futures

takezoe:
@longliveenduro Even if we use thread for asynchronous processing, we should use separated thread pools for CPU bound processing and I/O bound processing. The thread pool for CPU bound processing should have threads close to the number of CPU. One for I/O bound processing should have threads which are significantly larger than the number of CPUs.

Anyway, I think that allowing to give an execution context from user side is good strategy because it makes possible to control by users.

longliveenduro:
@takezoe yes exactly, seperate Threadpools for CPU and I/O bound blocking processing looks like a very good idea for a transition phase. Also looking at the Blogs they suggest a fixed Threadpool instead of a ForkJoin Threadpool (which the "standard" Scala thread pool is) for blocking I/O code.

Anyhow, looking at your PR, you are importing the standard Scala execution context in the methods that still use blocking via Await.result. What about using a separate Threadpool there whose size is either configurable or you pass it as a parameter. Passing as a parameter though is breaking the current contract and is a little bit strange since the method does not return an async instance (e.g. Future) which I would expect from a scala method which needs an ExecutionContext param.

Very interested what you think!

data/src/main/scala/org/apache/predictionio/data/store/LEventStore.scala
dszeto: LGTM. We can link to https://docs.scala-lang.org/overviews/core/futures.html#the-global-execution-context for a detail explanation if you'd like to.

dszeto:
How about we push for deprecating blocking methods? We can update docs to instruct developers to start using non-blocking methods instead.

takezoe:
I'm not sure if deprecating current blocking methods in the next version is good. I would like to see following work by @longliveenduro before making a decision.

takezoe:
Anyway, I'm going to merge. Thanks for all reviews!

PR #483, opened 2018-10-16
[PIO-184] livedoc 0.9.2 upgrade typo

Wei-1:
In this section:
https://github.com/apache/predictionio/blob/livedoc/docs/manual/source/resources/upgrade.html.md#2-in-xxxalgorithmscala

Line 136:
locate where `LEventStore.findByEntity()` is used, change it to `LEventStore.findByEntity()`

Should be changed to:
locate where `lEventsDb.findSingleEntity()` is used, change it to `LEventStore.findByEntity()`


dszeto:
Thanks!

PR #486, opened 2018-10-22
[PIO-187] Livedoc with Docker Installation Update

Wei-1:
After #462, we have Docker support in the repo.
This PR will modify the corresponding updates in Livedoc.

@marevol, please help to check if I make any mistake.


dszeto:
Thanks for writing this up! One small request: we have to declare that any Docker container image published to Docker Hub be **not official ASF releases**. Those can only be referred to as convenience binaries. We can say that the Docker build files in our Git repo as official though.

For more information please refer to http://www.apache.org/legal/release-policy.html.

Wei-1:
@dszeto, I will modify it for sure. thanks!

docs/manual/source/install/install-docker.html.md.erb
```diff
@@ -0,0 +1,127 @@
...
+
+```
+docker build -t predictionio/pio pio
+```
+
+One will be able to build an image with tag `prediction/pio:latest` using the above command.
+
+WARNING: People can get PredictionIO image from Dockerhub through `docker pull predictionio/pio`. However, since the image cannot run without a properly configured storage, please follow the following steps to complete the installation.
+
+WARNING: Image `prediction/pio` hosted on Dockerhub is **NOT** regarded as an official ASF release and might provide a different PredictionIO version from your desired PredictionIO version. It is recommended to build the image locally other than pulling directly from Dockerhub.
```
Wei-1: @dszeto how about this?

docs/manual/source/install/install-docker.html.md.erb
```diff
@@ -21,22 +21,34 @@ limitations under the License.
...
```
dszeto: We should tag Docker images with version as a separate work item.

dszeto:
LGTM. @marevol do you want to take a second look?

docs/manual/source/install/install-docker.html.md.erb
Wei-1: Agree. Currently, the image on Dockerhub is 0.12.1, which is different from the 0.13.0 that we are trying to target.

dszeto:
Going to merge this now. Thanks @Wei-1 !

@marevol if you see issues please open a separate ticket.

PR #487, opened 2018-10-23
[WIP][PIO-188] Update the build matrix on travis

shimamoto:
As for Scala, Spark and Hadoop, .travis.yml will cover all build patterns. As for storage modules, it will cover build patterns for each storage based on the default build.


shimamoto:
2 errored [Build #1609](https://travis-ci.org/apache/predictionio/builds/444959260)

It looks like ES1 storage module is not support EVENTDATA. I would like to correct my mistake. Whereas support for ES6 probably has bugs.

PR #489, opened 2018-10-26
[PIO-188] Update the build matrix on travis

shimamoto:
As for Scala, Spark and Hadoop, .travis.yml will cover all build patterns. As for storage modules, it will cover build patterns for each storage based on the default build.


takezoe:
LGTM. This policy makes sense and reducing test patterns would make Travis build more stable!

dszeto:
👍 Thanks for cleaning up the cross build @shimamoto !

PR #494, opened 2018-11-02
[PIO-192] Enhance PySpark support

shimamoto:
The limitations of the current Python support always force developers to have access to sbt. This enhancement will get rid of the build phase by implementing a default Python engine.

[PIO-192](https://issues.apache.org/jira/browse/PIO-192)


shimamoto:
@marevol @takezoe Can you take a look?

python/pypio/pypio.py
```diff
@@ -15,9 +15,35 @@
...
+    global sc
+    sc = spark.sparkContext
+    global sqlContext
+    sqlContext = spark._wrapped
+    global p_event_store
+    p_event_store = PEventStore(spark._jsparkSession, sqlContext)
+    print("Initialized pypio")
+
+
+def find(app_name):
```
takezoe: `findEvents` might be better to clarify the purpose of this function.

python/pypio/pypio.py
```diff
@@ -15,9 +15,35 @@
...
+    return p_event_store.find(app_name)
 
-p_event_store = PEventStore(spark._jsparkSession, sqlContext)
-cleanup_functions = CleanupFunctions(sqlContext)
 
+def save(model):
```
takezoe: `saveModel` might be better as same as above.

python/pypio/pypio.py
```diff
@@ -15,9 +15,35 @@
...
+    engine = sc._jvm.org.apache.predictionio.e2.engine.PythonEngine
+    engine.model().set(model._to_java())
+    main_args = utils.toJArray(sc._gateway, sc._gateway.jvm.String, sys.argv)
+    create_workflow = sc._jvm.org.apache.predictionio.workflow.CreateWorkflow
+    spark.stop()
+    create_workflow.main(main_args)
```
takezoe: This functions is wrapper of CreateWorkflow. It seems to can run even the evaluation by giving bootstrap arguments. Does it work?

python/pypio/pypio.py
shimamoto: Fixed it.

python/pypio/pypio.py
shimamoto: Fixed it.

python/pypio/pypio.py
shimamoto: Python code to specify for the --main-py-file option is only launched when kicking off a training.

python/pypio/pypio.py
```diff
@@ -15,9 +15,35 @@
...
+    global p_event_store
+    p_event_store = PEventStore(spark._jsparkSession, sqlContext)
+    print("Initialized pypio")
+
+
+def find(app_name):
+    return p_event_store.find(app_name)
 
-p_event_store = PEventStore(spark._jsparkSession, sqlContext)
-cleanup_functions = CleanupFunctions(sqlContext)
```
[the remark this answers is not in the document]
shimamoto: Currently, it looks like cleanup_functions is not referenced by anything. As for pyspark, let me check.

python/pypio/pypio.py
takezoe: Ah, top level commands don't accept illegal arguments so passing `sys.argv` directly to `CreateWorkflow.main()` is safe. I understand.

shimamoto:
@marevol @takezoe I updated the JIRA ticket. Again, could you please review this PR?

python/pypio/pypio.py
```diff
@@ -0,0 +1,125 @@
...
+import os
+import sys
+
+from pypio.data import PEventStore
+from pypio.utils import dict_to_scalamap, list_to_dict
+from pypio.workflow import CleanupFunctions
+from pyspark.sql import SparkSession
+
+
+def init(app_name=None):
```
takezoe: While I'm not sure the use case of pypio, I feel the lack of consistency in these APIs.

For example, `find_events()` takes `app_name`. Does this mean that pypio supports multiple apps in a same session? If so, Can we call `init()` multiple times to create multiple apps?

I think either of following options is good for API consistency:

- If pypio supports multiple apps, separate a method for creating a new app from `init()`.
- If pypio doesn't support multiple apps, make `app_name` mandatory in `init()` and use it for `find_events()` too.

shimamoto: You have a point. `init` function, as you say, shouldn't take `app_name`. I'll remove it.

Just an FYI, I found out the following:
- Running `pio app new <app_name>` creates an event table/index and an accesskey
- Running `pio accesskey new <app_name>` creates another accesskey for specified App
- Users must use an accessKey to import events via API

takezoe: Thanks for information! Actually, I had forgotten what `pio app new` (and `pio accessKey new`) do.

If you have a plan to support importing event data via Web API of the event server, how it handle accessKey can be an important point in the design of this Python support.

shimamoto: I'm considering supporting data import, but I don't have an image of how to implement yet. The ideal API specification is the following:

```
diabetes_df = h2o.import_file("../input/diabetes.csv", destination_frame="diabetes_df")
```

I believe we should see the exact behavior of current commands and SDK. Anyway, this topic is out of scope of my PR.

PR #495, opened 2018-11-06
[PIO-193] Async support for predict method and storage access, blocking code wrapped in blocking construct

longliveenduro:
See https://issues.apache.org/jira/browse/PIO-193
and 
https://lists.apache.org/thread.html/f14e4f8f29410e4585b3d8e9f646b88293a605f4716d3c4d60771854@%3Cuser.predictionio.apache.org%3E
and also https://jira.apache.org/jira/browse/PIO-182
for more details


takezoe:
I think this way isn't acceptable because it breaks current interfaces. This means that all existing templates won't work after this change. Async support in entire prediction service is really nice but we need to find another way.

longliveenduro:
@takezoe yes, you are definitely right. I'm just thinking about and trying some backward compatible tweaks. Will get back!

dszeto:
Thank you @longliveenduro ! Fingers crossed for your update. :)

longliveenduro:
@dszeto @takezoe This would be one possible solution to be nearly 100% backward compatible. With "nearly" I mean that an engine implementor has to add "override" to its implementation of predictBase() / predict() because in this suggestion I implement it with a default NotImplementedError stating that the method is deprecated.

If you want we could just drop this default implementation of BaseAlgorithm.predictBase / Subclass.predict and would have 100% compatibility but then even a new async implementation would have to implement these old method signatures with a dummy.

Other solution I did think of were using some runtime information (things like checking dynamically if a method is present or not) but I decided not to use that, because these kind of lookups is usually very slow and IMHO breaks type saftey.

PR #498, opened 2018-11-14
Improve readability of instructions

testflyjets:
Adds missing articles and tweaks the grammar.


shimamoto:
@testflyjets Thank you for your contribution! patch-2 branch is created from develop, so would you please create a branch from livedoc?

PR #500, opened 2018-11-19
[WIP] Incorporate the official templates in core repository

takezoe:
I'm considering about incorporate the official templates with main repository to reduce maintenance costs of them. My idea is:

- Check versions of PredictionIO/Scala/Spark/etc by TravisCI using a shell script and if these versions are not same among PredictionIO and templates, test will fail.
- Include the official templates and examples in the distribution which is made by `make-distribution.sh`.
- Documentations which refer these templates or examples will be fixed to be started from:
   ```
   $ cd $PIO_HOME/templates/template-scala-parallel-recommendation
   $ pio build
   ...
   ```
   instead of:
   ```
   $ git clone https://github.com/apache/predictionio-template-recommender.git
   $ cd predictionio-template-recommender
   $ pio build
   ...
   ```
Currently, for every release, we have to update versions of PredictionIO/Scala/Spark/etc, then create tag for each template repository by hand. If the official templates are in the same repository with PredictionIO itself, we will be able to finish this boring operation at once. Also, it will ensure that all official templates is synchronized with the latest PredictionIO version.

### DIFF
--- a/templates/template-scala-parallel-recommendation/LICENSE.txt
+++ b/templates/template-scala-parallel-recommendation/LICENSE.txt
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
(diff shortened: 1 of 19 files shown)


Wei-1:
I think it causes some confusions when new users try to learn PIO and found that they will need to clone the template from another repository.
So it is definitely better for us to put it together with the main repo.
Since we already have those examples in the main repo, I think it makes a lot of sense.

One question that I will bring up is:
Are we still going to use the package `org.example.recommendation`?
or maybe we could change it to `org.apache.predictionio.template`?

templates/template-scala-parallel-recommendation/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
takezoe: Is the license file necessary for each template after they are incorporated into this repository?

templates/template-scala-parallel-recommendation/data/import_eventserver.py
```diff
@@ -0,0 +1,56 @@
+"""
```
takezoe: Aren't Python file required license header?

templates/template-scala-parallel-recommendation/data/send_query.py
```diff
@@ -0,0 +1,7 @@
+"""
```
takezoe: Aren't Python files required license header?

templates/template-scala-parallel-recommendation/README.md
```diff
@@ -0,0 +1,79 @@
...
+limitations under the License.
+-->
+# Recommendation Template
+
+## Documentation
+
+Please refer to
+https://predictionio.apache.org/templates/recommendation/quickstart/.
+
+## Versions
```
takezoe: We will have to maintain README of each template after incorporating, I think it should be synchronized with PredictionIO release, but [text-classifier template](https://github.com/apache/predictionio-template-text-classifier) has inconsistent versioning with PredictionIO version. How should we handle that?

In addition, text-classifier template contains [large data](https://github.com/apache/predictionio-template-text-classifier/tree/develop/data) in its repository, I want to separate that data from the repository. Is there any suitable place to put them?

tests/unit.sh
```diff
@@ -43,4 +43,6 @@ sbt/sbt dataJdbc/compile test storage/test \
     -Delasticsearch.version=$PIO_ELASTICSEARCH_VERSION \
     -Dhbase.version=$PIO_HBASE_VERSION
 
+./tests/check_templates.sh
```
takezoe: Add incorporated templates here to check by TravisCI.

tests/check_templates.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
takezoe: This script is run after unit tests on TravisCI to check version consistency and test templates.

tests/check_templates.sh
```diff
@@ -0,0 +1,57 @@
...
+cd ${FWDIR}
+
+# Extract information from PIO's build configuration
+PIO_VERSION=$(grep ^version ${FWDIR}/build.sbt | grep ThisBuild | grep -o '".*"' | sed 's/"//g')
+SCALA_VERSION=$(grep ^scalaVersion ${FWDIR}/build.sbt | grep ThisBuild | grep -o ', ".*"' | sed 's/[, "]//g')
+
+echo "======================================================================="
+echo "PIO_VERSION: $PIO_VERSION"
+echo "SCALA_VERSION: $SCALA_VERSION"
+echo "======================================================================="
```
takezoe: TODO: Check Spark version

templates/template-scala-parallel-recommendation/README.md
Wei-1: I think the test you added to `check_templates.sh` is a good idea.
People will know that they have to change the corresponding template version while modifying PIO's version.
As for the data, maybe we add another non-test branch?

templates/template-scala-parallel-recommendation/build.sbt
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+name := "template-scala-parallel-recommendation"
+
+scalaVersion := "2.11.12"
+libraryDependencies ++= Seq(
+  "org.apache.predictionio" %% "apache-predictionio-core" % "0.14.0-SNAPSHOT" % "provided",
+  "org.apache.spark"        %% "spark-mllib"              % "2.1.1" % "provided")
```
Wei-1: Based on Travis-CI build test: https://travis-ci.org/apache/predictionio/builds/457523767
Will need to change from `2.1.1` to `2.1.3`

takezoe: Updated. Test failed correctly. :-)

tests/check_templates.sh
takezoe: Added Spark(MLLib) version checking

PR #503, opened 2018-12-13
[WIP][PIO-199] Spark 2.4 support for Scala 2.11

shimamoto:
Though Spark 2.4 has added Scala 2.12 support, it is marked as experimental on the Spark release notes, so the next version of PIO will add Spark 2.4 support for Scala 2.11 only.


shimamoto:
@takezoe @dszeto Can you take a look?

shimamoto:
@takezoe Thanks for review! In terms of Spark 2.2.3, would it be better if I create another PR?

PR #507, opened 2019-01-17
[PIO-203] Fixes pio status warnings in ES storage

shimamoto:
ref #466

This PR includes several fixes as below:

1. Fixes a bug that unnecessary indices are created
`null_sequences` is clearly incorrect.

1. Remove the `_all` part of each mapping
[_all] is deprecated in 6.0+ and will be removed in 7.0. As a replacement, you can use [copy_to] on mapping fields to create your own catch all field.

1. The number of shards/replicas
We would recommend defining Index templates (https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html) if a user want to add specific settings associated with each index. I would handle it later.


EmergentOrder:
LGTM

shimamoto:
@EmergentOrder @marevol Thanks for review!

PR #514, opened 2019-05-16
Fixed docker's build failure

azliabdullah:
The docker build was failing at step 19/24 due to spark's download URL [original URL](http://www-us.apache.org/dist/spark/spark-2.2.3/spark-2.2.3-bin-hadoop2.7.tgz) hit HTTP 404 error.

I have updated the link to point to [archive URL](https://archive.apache.org/dist/spark/spark-2.2.3/spark-2.2.3-bin-hadoop2.7.tgz) which contains correct location of the file. And also safer to use as its housing old versions and latest versions as well.


shimamoto:
@azliabdullah Thanks!

PR #519, opened 2019-06-26
Update enginevariants.html.md

luto65:
added description of the commandline and examples


Wei-1 (commented):
Maybe this will make more sense?
```
In order to deploy multiple variants,
one must specify a JSON file containing different variants,
a port,
and an instance-id to an engine.
At the end of the training,
the instance-id will be stored in the `````id````` column
of the `````pio_meta_engineinstances````` table.
```

PR #533, opened 2020-12-29
Release/0.14.0

vking34:


docker/pio/Dockerfile
```diff
@@ -17,8 +17,8 @@
 
 FROM openjdk:8
 
-ARG PIO_GIT_URL=https://github.com/apache/predictionio.git
-ARG PIO_TAG=v0.13.0
+ARG PIO_GIT_URL=https://github.com/vking34/predictionio.git
```
Wei-1, line 20: This should stay as `apache` right?

